PR #32, opened 2026-03-13
refactor: extract display engine into display.ts

jasoncrawford:
Pure refactor — no behavior changes. Moves the display engine (colors, formatting, markdown, printing, status line, VERBOSE flag) from src/repl.ts into a new src/display.ts module. src/repl.ts re-exports everything so existing imports and tests are unaffected.

### DIFF
--- a/src/repl.ts
+++ b/src/repl.ts
@@ -1,12 +1,7 @@
 import fs from "fs";
 import { fileURLToPath } from "url";
 import { query, type HookCallback } from "@anthropic-ai/claude-agent-sdk";
-import {
-  W, hr, VERBOSE, setVerbose, c, s,
-  trunc, fmtCount, fmtStats, fmtArgs,
-  startStatus, stopStatus, print,
-  toolUseNames, printMessage, printHook,
-} from "./display.js";
+import * as display from "./display.js";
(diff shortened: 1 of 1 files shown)


src/repl.ts
```diff
@@ -1,6 +1,13 @@
 import fs from "fs";
 import { fileURLToPath } from "url";
 import { query, type HookCallback } from "@anthropic-ai/claude-agent-sdk";
+import {
+  W, hr, VERBOSE, setVerbose, c, s,
+  trunc, fmtCount, fmtStats, fmtArgs,
+  startStatus, stopStatus, print,
+  toolUseNames, printMessage, printHook,
+} from "./display.js";
+export * from "./display.js";
```
jasoncrawford: There's still a lot of imports here, and also an `export *`. Just, wouldn't it be cleaner to import `display` as a module, and then invoke, e.g., `display.setVerbose`, `display.fmtCount`, etc.?

jasoncrawford: Done — switched to `import * as display from "./display.js"` and prefixed all call sites with `display.`. The `export * from "./display.js"` line stays (one line) so tests importing display symbols from `repl.js` continue to work without changes.

Issue #96, opened 2026-03-16
Worker output is double-spaced


jasoncrawford:
Excerpt from test logs below. There's an extra newline in between each piece of output, although not within multi-line chunks. This is probably related to recent console/terminal changes.

```
Now let me read the relevant source files.


• Glob(src/**/*.ts)


• Glob(tests/**/*.ts)

→ /workspace/src/types.ts /workspace/src/github.ts /workspace/src/worker.ts /workspace/src/display.ts…

→ /workspace/tests/helpers.ts /workspace/tests/fixtures/messages.ts /workspace/tests/foreman.registry…


• Read(/workspace/src/templates.ts)

→ 71 lines


• Read(/workspace/tests/templates.test.ts)

→ 60 lines


• Read(/workspace/tests/worker.test.ts)

→ 276 lines


I can see that tests/templates.test.ts already exists with some tests! But looking at the issue description, the existing tests are incomplete compared to what the issue requires. Let me compare what exists vs. what the issue asks for:

Existing tests:
• buildInitialPrompt: includes number/title/body, handles empty body
• buildEventPrompt: multi-event fallback, check_run failure, pull_request_review, unknown event fallback

What the issue asks for that's missing:
• buildInitialPrompt: includes labels, repoUrl; handles null body; handles empty labels array; handles multiple labels

```

PR #101, opened 2026-03-17
fix: clear input print callback during runQuery to fix double-spacing (#96)

jasoncrawford:
## Summary

- In worker mode, `ask()` registers `drawFresh()` as `_inputPrintCallback` so the prompt redraws when background WebSocket messages arrive while the user is waiting at the prompt
- While a query is running (`runQuery`), the worker's `ask()` loop is also active — meaning `drawFresh()` fires after **every** `display.print()` call, writing an extra `\r\n` before re-rendering the prompt
- This extra `\r\n` is the double-spacing observed in the issue (introduced by the `_inputPrintCallback` mechanism added in the multiline input fix, commit `9c03aed`)

**Fix:** Call `display.setInputPrintCallback(null)` at the start of `runQuery`. This suppresses `drawFresh()` during query output. `ask()` re-registers the callback on each new call, so prompt-redrawing after background notifications still works normally between query runs.

## Test plan

- [x] New test in `repl.query.test.ts`: sets a mock callback via `setInputPrintCallback`, runs `runQuery`, verifies the callback was **not** called during output
- [x] All 3079 existing tests pass

### DIFF
--- a/src/display.ts
+++ b/src/display.ts
@@ -309,7 +309,12 @@ export function print(line: string | null) {
   _clearStatus();
   console.log(line);
   _drawStatus();
-  _inputPrintCallback?.();
+  // Only redraw the input prompt when no query is running. During a query the
+  // status bar is active; calling drawFresh() then would interleave the prompt
+  // with query output and corrupt the display (causing double-spacing and
+  // swallowed output). ask() re-registers the callback on each new call, so
+  // prompt-redrawing after background notifications still works between runs.
+  if (!_statusActive) _inputPrintCallback?.();
 }
 
 export function resolve(table: FmtTable, key: string, data: any): string | null {
--- a/src/input.ts
+++ b/src/input.ts
@@ -427,13 +427,16 @@ export function ask(
     function submit(value: string) {
       if (done) return;
       done = true;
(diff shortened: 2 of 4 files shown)


jasoncrawford:
Something funky is still going on. There is still a double-space after the prompt, only. But also, some console printing seems to be getting swallowed. Note in the below, there's no printing of a git command before answering the branch.

```
> What git branch are we on?


issue-96-worker-double-spacing

2s, 1 turn, tokens: 3 in / 14 out

> What files do you see?


The user is asking what files are in the workspace. Let me check for any applicable skills first - this is a simple informational question about listing files, so I don't think a specific skill applies here beyond the general "using-superpowers" one already loaded.

$ ls /workspace # List files in workspace
→ CLAUDE.md bin docs eslint.config.js node_modules package-lock.json package.json repl.log src tech-n…

  CLAUDE.md
  bin/
  docs/
  eslint.config.js
  node_modules/
  package-lock.json
  package.json
  repl.log
  src/
  tech-notes.md
  tests/
  tsconfig.json
  vitest.config.ts

8s, 2 turns, tokens: 4 in / 212 out
```

PR #141, opened 2026-03-18
feat: better command autocomplete

### LINKED ISSUE #125: Better command autocomplete
Right now, up to three slash-commands are autocompleted, and they are shown inline.

We should:
* show up to 5 commands
* one per line
* next to each command, show its description, or if it doesn't have a description, the first line of its text; truncated
* show all the descriptions left-aligned with each other

Here's an example of how Claude Code does it, for inspiration.

```

────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
❯ / 
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
  /branch-discipline                  Use when editing files in projects that have CI or PRs set up, or otherwise use branches, or whenever multiple agents might work in parallel (user)     
  /notes-before-clear                 Use before running /clear to save session notes to memory and project docs (user)
  /devcontainer-setup                 Use when setting up a new project for containerized Claude Code development, when the user wants to run Claude with --dangerously-skip-permissions safe…
  /setting-up-ci                      Use when starting a new project, adding tests to a project without CI, or when a project has tests that only run locally (user)
  /loop                               Run a prompt or slash command on a recurring interval (e.g. /loop 5m /foo, defaults to 10m) (bundled)
  /add-dir                            Add a new working directory
```

jasoncrawford:
## Summary

- Show up to 5 suggestions (was 3), one per line instead of inline
- Show description next to each command, left-aligned across all displayed suggestions
- Description is sourced from the `description` frontmatter field of the command/skill file, or the first non-empty line of the file content as fallback
- Built-in commands (clear, exit) show no description

## Implementation details

- New exported `CommandSuggestion` type: `{ name: string; description: string }`
- New `listCommands()` and `listWorkerCommands()` functions return `CommandSuggestion[]` (replacing `listCommandNames`/`listWorkerCommandNames` as the input to `ask()`)
- `ask()` now accepts `() => CommandSuggestion[]` as its `getCommands` parameter
- `renderSuggestions()` pads command names so all descriptions start at the same column; truncates descriptions to fit terminal width
- Submit now uses `\x1b[J` (erase to end of screen) instead of `\x1b[K` (erase to end of line) to correctly clear all suggestion rows at once

Closes #125

### DIFF
--- a/src/input.ts
+++ b/src/input.ts
@@ -47,14 +47,15 @@ export type SlashCommandResult =
 
 type BuiltinCommand = {
   name: string;
+  description: string;
   result: Exclude<SlashCommandResult, { type: "unknown_command" }>;
   workerOnly?: boolean;
 };
 
 const BUILTIN_COMMANDS: BuiltinCommand[] = [
-  { name: "clear",         result: { type: "clear" } },
-  { name: "exit",          result: { type: "exit" } },
-  { name: "task-complete", result: { type: "task_complete" }, workerOnly: true },
+  { name: "clear",         description: "Clear the conversation",      result: { type: "clear" } },
+  { name: "exit",          description: "Exit the REPL",               result: { type: "exit" } },
+  { name: "task-complete", description: "Mark the current task as done", result: { type: "task_complete" }, workerOnly: true },
 ];
 
 /**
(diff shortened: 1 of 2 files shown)


jasoncrawford:
Nice, but builtin commands are missing descriptions. Let's give them some simple descriptions

PR #221, opened 2026-03-22
fix: bidirectional depsLoaded sync in reconcile() to prevent stale-dep assignment

### LINKED ISSUE #220: bug: edited handler doesn't prevent stale-dep assignment during dep reload
## Background

PR #218 (refactor: reactive task queue) simplified the `edited` handler to:

```typescript
if (action === "edited") {
  const changes = p.changes as Record<string, unknown> | undefined;
  if (changes?.body) {
    const newBody = String(issue.body ?? "");
    const entry = labeledIssues.get(issueNumber);
    if (entry) {
      entry.depsLoaded = false;      // resets canonical state
      entry.issue = { ...entry.issue, body: newBody };
      startDepsLoad(issueNumber, newBody);  // async; calls reconcile() when done
    }
  }
  // fall through: let forwardEvent run for assigned tasks
}
```

## Problem

`reconcile()` step 2 only syncs `depsLoaded` in one direction — `false → true`:

```typescript
for (const [num, { depsLoaded }] of labeledIssues) {
  if (depsLoaded) {   // ← never propagates false
    const t = taskQueue.getTaskForIssue(num);
    if (t && !t.depsLoaded) taskQueue.markDepsLoaded([num]);
  }
}
```

So after `edited` fires:
- `labeledIssues.get(num).depsLoaded === false` ✓
- `task.depsLoaded === true` ✗ (not propagated)

If any other event triggers `reconcile()` before `startDepsLoad` completes (e.g. a worker connecting, an issue reopening), `tryAssignWork` sees `task.depsLoaded === true` and can assign the task with stale blocker data.

## Fix

Make step 2 a full bidirectional sync:

```typescript
// Step 2: sync depsLoaded from labeledIssues to existing tasks (both directions)
for (const [num, { depsLoaded }] of labeledIssues) {
  const t = taskQueue.getTaskForIssue(num);
  if (t && t.depsLoaded !== depsLoaded) {
    if (depsLoaded) taskQueue.markDepsLoaded([num]);
    else t.depsLoaded = false;  // or a new taskQueue.unloadDeps() method
  }
}
```

This makes the model clean: `edited` sets `entry.depsLoaded = false`, the next `reconcile()` (from any trigger) propagates it to the task making it non-assignable, and when `startDepsLoad` finishes it sets `entry.depsLoaded = true` and calls `reconcile()` which propagates `true` back — task becomes assignable with fresh dep data. "Load all the data, then reconcile."

jasoncrawford:
## Summary

- `reconcile()` step 2 previously only synced `depsLoaded` in one direction (`false → true`), leaving `task.depsLoaded = true` when `labeledIssues.get(n).depsLoaded` was reset to `false` by the `edited` handler
- Any subsequent `reconcile()` call (worker connecting, issue reopening, etc.) before `startDepsLoad` completed could assign the task with stale blocker data
- Made step 2 a full bidirectional sync: now propagates `false` to the task too, making it non-assignable until `startDepsLoad` resolves and sets `depsLoaded = true`

## Test plan

- [x] Added test: `syncs depsLoaded=false from labeledIssues to an existing task that has depsLoaded: true` — verifies the false propagation
- [x] Added test: `does not assign a pending task via tryAssignWork when its depsLoaded is false after reconcile` — verifies a task with stale depsLoaded is not assigned to an idle worker
- [x] All 781 tests pass

Closes #220

### DIFF
--- a/src/foreman.ts
+++ b/src/foreman.ts
@@ -789,10 +789,7 @@ export function createForemanWss(
     // Step 2: sync depsLoaded from labeledIssues to existing tasks (both directions)
     for (const [num, { depsLoaded }] of labeledIssues) {
       const t = taskQueue.getTaskForIssue(num);
-      if (t && t.depsLoaded !== depsLoaded) {
-        if (depsLoaded) taskQueue.markDepsLoaded([num]);
-        else t.depsLoaded = false;
-      }
+      if (t) t.depsLoaded = depsLoaded;
     }
 
     // Step 3: remove pending tasks whose issue no longer has the label


src/foreman.ts
```diff
@@ -786,13 +786,12 @@ export function createForemanWss(
...
-      if (depsLoaded) {
-        const t = taskQueue.getTaskForIssue(num);
-        if (t && !t.depsLoaded) {
-          taskQueue.markDepsLoaded([num]);
-        }
+      const t = taskQueue.getTaskForIssue(num);
+      if (t && t.depsLoaded !== depsLoaded) {
+        if (depsLoaded) taskQueue.markDepsLoaded([num]);
+        else t.depsLoaded = false;
       }
```
jasoncrawford: why is this not as simple as

```
if (t) t.depsLoaded = depsLoaded
```

what is markDepsLoaded() doing that is special, and why is there no equivalent for marking false?

jasoncrawford: Good catch. `markDepsLoaded` just does `t.depsLoaded = true` — nothing more. The asymmetry was needless. Simplified to `if (t) t.depsLoaded = depsLoaded` in 58e1115.

PR #294, opened 2026-03-27
feat: prepend timestamps to display.print() output in verbose mode

### LINKED ISSUE #279: diagnostics: prepend timestamps to display.print() output in verbose mode
## Problem

Worker output via `display.print()` has no timestamps. When diagnosing disconnection behaviour it's impossible to tell from the log when a disconnect happened relative to other activity (task start, event receipt, query completion, etc.).

The foreman already prepends ISO timestamps to every line via `flog()`. Workers have no equivalent.

## What to change

In `src/display.ts`, in the `print()` function, prepend `fmtTime()` (or `fmtTimestamp()` for ISO format) to each line when `verbose` is true:

```typescript
export function print(line: string | null) {
  // ...
  const prefix = verbose ? c.darkGray(fmtTime() + " ") : "";
  // prepend prefix to the output line before writing
}
```

`fmtTime()` returns `HH:MM:SS` which is compact and human-readable in a terminal. `fmtTimestamp()` returns full ISO which is better for log aggregation and cross-referencing with the DB. Either works — pick whichever fits best.

## Why verbose mode

In non-verbose mode the output is already fairly dense and timestamps on every line would add noise. In verbose mode the user has opted into detailed output, so timestamps are a natural fit.

## Why this helps

Timestamps on worker output let you correlate:
- When a disconnect happened vs. when a task was running
- Whether events arrived before or after a reconnect
- How long reconnects take in practice

Part of the Stage 1 diagnostics work: #276 #277 #278.

jasoncrawford:
## Summary

- In `print()`, prepend `c.darkGray(fmtTime() + " ")` to each line when `verbose` is `true`
- No-op when verbose is `false`, keeping non-verbose output unchanged
- Uses compact `HH:MM:SS` format matching the issue recommendation

## Test plan

- [ ] New test file `tests/display.print-timestamps.test.ts` covers: no timestamp in non-verbose mode, timestamp present and styled in darkGray in verbose mode, timestamp appears before message content, `print(null)` still no-ops in verbose mode
- [ ] All 901 existing tests continue to pass

Closes #279

### DIFF
--- a/src/display.ts
+++ b/src/display.ts
@@ -566,8 +566,14 @@ export function print(line: string | null) {
   // prompt). The prompt is then redrawn below via _inputPrintCallback.
   if (_inputPrintCallback) process.stdout.write("\r\x1b[K");
   if (verbose) {
-    const ts = c.darkGray(fmtTime() + " ");
-    console.log(line.split("\n").map(l => ts + l).join("\n"));
+    // \x1b[39m resets only the foreground ("pop-color") rather than \x1b[0m
+    // which would reset everything including the content's own color codes.
+    const ts = `\x1b[90m${fmtTime()} \x1b[39m`;
+    const parts = line.split("\n");
+    // Re-apply any opening ANSI color codes to continuation lines so that
+    // inserting the timestamp prefix doesn't strip color mid-block.
+    const openColor = (line.match(/^(\x1b\[[0-9;]*m)+/) ?? [""])[0];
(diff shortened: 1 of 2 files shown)


jasoncrawford:
I like this formatting better, but it loses the coloring on the output. Can the grey timestamp have like a pop-color end code rather than a reset-to-white end code?

PR #310, opened 2026-03-28
feat: disconnected worker state — keep registry entry and queue events on disconnect

### LINKED ISSUE #306: Foreman: disconnected worker state — keep registry entry and queue events on disconnect
Currently when a worker's WebSocket closes, `registry.remove(workerId)` is called immediately. Any GitHub events that arrive during the reconnect window hit `routeEvent()`, see the task as `assigned`, try `registry.send()`, find no worker, and are silently dropped. The worker reconnects to silence, missing CI results, review comments, etc.

### New `"disconnected"` status in WorkerRegistry

On WebSocket close:
- **Worker had no task (was idle):** remove from registry as today
- **Worker had an assigned task:** mark status as `"disconnected"`, record `disconnectedAt`, keep the task `assigned`

### Event queuing for disconnected workers

In `routeEvent()`, when the assigned worker's status is `"disconnected"`, queue the event on the task (same `pendingEvents` mechanism already used for `pending` tasks) instead of dropping it.

### On reconnect with busy status

Worker sends `worker_hello { status: busy, taskId }`. Foreman finds the disconnected registry entry:
- Updates the WebSocket reference, sets status back to `"busy"`
- Cancels the reclaim timer (see #308)
- Drains queued events and forwards them

### On reconnect with idle status

If a worker sends `worker_hello { status: idle }` but the foreman has an outstanding disconnected entry for it (with an assigned task), the foreman should:
- Cancel the reclaim timer
- Revert the task to `"pending"`
- Register the worker as idle normally

This covers the case where the worker process died and restarted — it has no session context and can't resume. There is no such thing as resuming a crashed worker's session. Treat it as a new idle worker.

This rule applies whether the disconnected entry came from a mid-session drop (runtime) or was pre-populated from the DB on startup (see #309).

## What changes

- `WorkerRegistry`: add `"disconnected"` status, `disconnectedAt` field
- `foreman.ts` `ws.on("close")`: branch on whether worker had a task
- `routeEvent()`: queue event on task when assigned worker is `"disconnected"`
- `worker_hello` handler: handle reconnect to disconnected registry entry (busy → resume, drain events; idle → cancel timer, revert task to pending)

## Extended by

- #309 adds startup pre-population of disconnected entries from the DB, so this in-memory behaviour also kicks in correctly after a foreman restart

jasoncrawford:
## Summary

- Adds `"disconnected"` status to `WorkerState` in `WorkerRegistry`, with a `disconnectedAt` timestamp
- On WebSocket close: if the worker had an active task, call `markDisconnected()` (keep registry entry); otherwise remove as before
- In `forwardEvent()`: if the assigned worker is `"disconnected"`, queue the event on the task instead of dropping it
- On `worker_hello` reconnect with `status: busy`: existing reclaim logic handles it — registry entry is replaced, queued events are drained and forwarded
- On `worker_hello` reconnect with `status: idle`: detects the disconnected entry, calls new `taskQueue.revertTask()` to put the task back to `"pending"`, then registers the worker idle normally
- Updates `WorkerSnapshot` type in `admin-ws.ts` to include `"disconnected"` status

## Test plan

- [ ] `WorkerRegistry.markDisconnected` sets status, keeps taskId, records timestamp
- [ ] `getIdleWorker` excludes disconnected workers
- [ ] Worker with active task is marked `disconnected` (not removed) on WebSocket close
- [ ] Idle worker is still removed on close
- [ ] Events are queued (not dropped) when assigned worker is disconnected
- [ ] Worker reconnecting as `busy` drains queued events
- [ ] Worker reconnecting as `idle` reverts task to pending and gets reassigned
- [ ] All 924 existing tests continue to pass

Closes #306

### DIFF
--- a/src/foreman.ts
+++ b/src/foreman.ts
@@ -762,11 +762,10 @@ export function createForemanWss(
           log(workerId, "→ standby");
         }
       } else {
-        // If there's a disconnected entry for this worker with an assigned task,
-        // the worker process restarted and can't resume. Revert the task to pending.
+        // If this worker had an assigned task, it can't resume — revert to pending.
         const existing = registry.get(workerId);
-        if (existing?.status === "disconnected" && existing.currentTaskId) {
-          log(workerId, `hello idle (was disconnected on task #${existing.currentTaskId}) — reverting task to pending`);
+        if (existing?.currentTaskId) {
+          log(workerId, `hello idle (had task #${existing.currentTaskId}) — reverting task to pending`);
           taskQueue.revertTask(existing.currentTaskId);
         } else {
           log(workerId, "hello idle");


src/foreman.ts
```diff
@@ -447,11 +462,16 @@ export function createForemanWss(
...
-        flog(`[task ${ref}] ${evt.name} DROPPED — worker ${task.assignedWorkerId.slice(0, 8)} not in registry (disconnected?)`);
+      if (worker?.status === "disconnected") {
+        taskQueue.queueEvent(task.taskId, evt);
+        flog(`[task ${ref}] ${evt.name} queued (worker ${task.assignedWorkerId.slice(0, 8)} disconnected)`);
+      } else {
+        if (!worker) {
+          flog(`[task ${ref}] ${evt.name} DROPPED — worker ${task.assignedWorkerId.slice(0, 8)} not in registry (disconnected?)`);
+        }
+        registry.send(task.assignedWorkerId, { type: "event_notification", taskId: task.taskId, event: evt });
+        log(task.assignedWorkerId, `→ event_notification ${ref} ${evt.name}`);
```
jasoncrawford: I know this is pre-existing behavior, but I'm unclear what's happening here... we do these lines even if the worker doesn't exist?

PR #333, opened 2026-03-28
refactor: DRY up tryAssignWork + broadcastSnapshot in WS message handlers

### LINKED ISSUE #317: Foreman: DRY up tryAssignWork + broadcastSnapshot at end of message handlers
## Background

Most (possibly all) foreman WebSocket message handlers end with the same two calls:

```typescript
tryAssignWork();
broadcastSnapshot();
```

This was noted during review of #314 as repeated boilerplate that could be cleaned up.

## Proposal

Route all message handling through a common exit path that automatically calls `tryAssignWork()` and `broadcastSnapshot()` after each handler, rather than repeating it in every handler individually.

Closes #308 (tracked in PR #314)

jasoncrawford:
## Summary

- Removes the repeated `broadcastSnapshot()` + `tryAssignWork()` boilerplate that appeared at the end of every WebSocket message handler (`handleWorkerHello`, `handleTaskComplete`, `handleWorkerGoodbye`)
- The message dispatch in `ws.on("message", ...)` now calls `assignIdleWorkers()` after each recognized handler — this function already encapsulates both calls
- One intermediate `broadcastSnapshot()` in `handleWorkerHello`'s "reclaimed" branch is preserved since it announces task assignment state before draining queued events

## Test plan

- [x] All 977 existing tests pass — no behavior change
- [x] Type check passes (`npx tsc --noEmit`)

Closes #317

### DIFF
--- a/src/foreman.ts
+++ b/src/foreman.ts
@@ -925,10 +925,11 @@ export function createForemanWss(
       });
       broadcastMessageEvent({ direction: "received", workerId: rcvWorkerId, taskId: rcvTaskId, msgType: msg.type });
 
-      if (msg.type === "worker_hello") { handleWorkerHello(msg); assignIdleWorkers(); }
-      else if (msg.type === "task_complete") { handleTaskComplete(msg); assignIdleWorkers(); }
-      else if (msg.type === "worker_goodbye") { handleWorkerGoodbye(msg); assignIdleWorkers(); }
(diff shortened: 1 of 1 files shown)


jasoncrawford (commented):
Good catch — moved `assignIdleWorkers()` to a single call at the bottom of the dispatch, using `return` on the unknown-type branch to skip it.

PR #345, opened 2026-03-28
fix: log event_notification messages forwarded to active workers

### LINKED ISSUE #341: Missing webhook event messages to worker
The messages that the foreman sends to the worker to forward webhooks are not showing up in any of the event logs.

Since workers detail pages only show messages and not webhooks, that means they don't show any of the webhook events at all.

The webhook forwarding messages should be in the event logs.

jasoncrawford:
## Summary

- `forwardEvent` in the foreman was sending `event_notification` messages to active workers without calling `dbLogger.logForemanMessage` or `broadcastMessageEvent`
- This meant these messages never appeared in the DB `foreman_messages` table, so `queryWorkerMessages` (used by worker detail pages) returned nothing for webhook-forwarding events
- The admin real-time event log was also missing these entries

## Fix

In the `else if (worker)` branch of `forwardEvent`, added the same `dbLogger?.logForemanMessage(...)` and `broadcastMessageEvent(...)` calls that already exist when draining queued events during task assignment.

## Tests

Added `tests/foreman.event-forwarding-logging.test.ts` with 5 tests verifying:
- `logForemanMessage` is called with `event_notification` type, correct `workerId`, `taskId`, and `direction: "sent"`
- `broadcastLogEvent` is called with a `kind: "message"` entry including `event_notification` summary and correct `taskId`/`workerId`
- Both issue_comment and check_suite event types are covered

Closes #341

### DIFF
--- a/src/foreman.ts
+++ b/src/foreman.ts
@@ -503,7 +503,8 @@ export function createForemanWss(
     });
   }
 
-  function sendMsg(workerId: string, taskId: string | null, msg: ForemanMessage): void {
+  function sendMsg(workerId: string, msg: ForemanMessage): void {
+    const taskId = "taskId" in msg ? msg.taskId : null;
     registry.send(workerId, msg);
     dbLogger?.logForemanMessage({ direction: "sent", workerId, taskId, msgType: msg.type, payload: msg as unknown as Record<string, unknown> });
     broadcastMessageEvent({ direction: "sent", workerId, taskId, msgType: msg.type });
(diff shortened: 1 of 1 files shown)


jasoncrawford (commented):
Is there an error-prone coupling here, that every time we call registry.send, we have to remember to call dbLogger.logForemanMessage? Can we write the code to eliminate that error or make it harder to commit, without violating good separation of concerns?

src/foreman.ts
```diff
@@ -503,6 +503,12 @@ export function createForemanWss(
     });
   }
 
+  function sendMsg(workerId: string, taskId: string | null, msg: ForemanMessage): void {
```
jasoncrawford: is taskId always the same as msg.taskId? if so we don't have to pass it in separately?

PR #347, opened 2026-03-28
fix: send recent events to new dashboard clients via WebSocket buffer

### LINKED ISSUE #339: Recent events on dashboard keep getting reset
I thought this was fixed by a previous issue, #249? But the dashboard still shows no events except ones that came in while the page was open. Whereas they are all available on the event log. I think the dashboard should just show the last ~30 events from the log or something

jasoncrawford:
## Summary

- The dashboard's Recent Events section was empty on load because it relied on a separate `/api/log` HTTP fetch that fails silently when the DB is not configured
- Added an in-memory ring buffer (last 30 entries) to `createAdminWss` in `src/admin-ws.ts`
- When a new WebSocket client connects, it now receives an `initial_log` message with buffered recent events (after the snapshot)
- Dashboard handles the new `initial_log` message type to populate Recent Events, replacing the unreliable fetch

## Test plan

- [ ] All 993 tests pass (`npm test`)
- [ ] `tests/admin-ws.test.ts`: 4 new tests verify buffer behavior (cap at 30, newest-first, sent on connect, ordering after snapshot)
- [ ] `frontend/tests/Dashboard.test.tsx`: updated to test `initial_log` handling instead of `/api/log` fetch

Closes #339

### DIFF
--- a/frontend/src/pages/Dashboard.tsx
+++ b/frontend/src/pages/Dashboard.tsx
@@ -1,4 +1,4 @@
-import { useState, useEffect, useCallback } from "react";
+import { useState, useCallback } from "react";
 import { Link } from "react-router-dom";
 import { useAdminWs } from "../hooks/useAdminWs.ts";
 import type { TaskSnapshot, WorkerSnapshot, LogEntry, AdminMessage } from "../types.ts";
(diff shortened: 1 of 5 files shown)


jasoncrawford:
Wait… db is configured in prod, so I don't think this is the cause of the issue I saw.

jasoncrawford:
Good catch. The DB explanation in the PR description was overly narrow — the real issue is that the separate HTTP fetch approach in Dashboard.tsx was inherently fragile in multiple ways:

1. **Race condition with live events**: When the dashboard mounts, any `log_event` WebSocket messages that arrive *before* the `/api/log` fetch completes get prepended to the empty state. When the fetch then resolves, `setRecentLog(historicalEntries)` **replaces** the array, silently discarding those already-received live events. The user would see events appear, then vanish, then reappear as new events come in — which matches "keep getting reset".

2. **Navigation resets**: Each time the user navigates away and back, the component remounts with an empty `recentLog`. The fetch refires, but until it resolves the section shows nothing (and may flicker as above).

The WebSocket buffer fix avoids all of this: the `initial_log` message is delivered synchronously on connect (before any live `log_event` messages can arrive), so there's no window where state gets replaced. Does that match the behavior you were seeing?

Issue #312, opened 2026-03-28
Worker/Foreman: handshake lifecycle — worker waits for hello_ack before sending task messages


jasoncrawford:
Currently a worker sends `worker_hello` on WebSocket open and immediately considers itself connected. If the worker was mid-task, it may attempt to send `task_complete` before the foreman has processed the hello and determined whether the task is still assigned to this worker. This creates a window where the worker sends task messages the foreman is not ready to interpret correctly.

### New foreman message: `hello_ack`

The foreman sends `hello_ack` in response to every `worker_hello`, after processing it:

```typescript
{ type: "hello_ack", workerId: string, status: "idle" | "busy" | "cancelled" }
```

- `"busy"` — task reclaimed, worker may resume (queued events follow immediately after)
- `"idle"` — worker registered as idle, may receive `task_assigned` shortly
- `"cancelled"` — task was reassigned while worker was disconnected; worker should stop and reset (equivalent to `task_cancelled`, see #301)

### Worker lifecycle

The worker gains a connection state: `connecting` → `hello_sent` → `registered`.

- In `hello_sent` state: worker buffers any outgoing task messages (e.g. `task_complete`)
- On receiving `hello_ack`: transition to `registered`, flush buffered messages (unless status is `"cancelled"`, in which case discard them and stop)

This way a `task_complete` that was queued while the WebSocket was down is only sent after the handshake confirms the worker still owns the task.

### Foreman: always verify worker ownership on task messages

Regardless of lifecycle, the foreman should validate that the sender owns the task before acting on `task_complete` (and any other task-scoped message). If `task.assignedWorkerId !== workerId`, ignore the message. This is a belt-and-suspenders check that should hold even without the handshake lifecycle change.

## What changes

- `types.ts`: add `hello_ack` to `ForemanMessage`
- `foreman.ts` `worker_hello` handler: send `hello_ack` after processing; add worker ID check to `task_complete` handler (and any other task-scoped message handlers)
- `worker.ts`: add `registered` state; buffer outgoing task messages until `hello_ack` received; handle `hello_ack` with `"cancelled"` status as equivalent to `task_cancelled`

## Depends on

- #306 (disconnected worker state — the reconnect scenario this is designed around)
- #301 (`task_cancelled` — `hello_ack` with status `"cancelled"` replaces the need to send a separate `task_cancelled` on hello)

PR #351, opened 2026-03-29
refactor: remove standby message from protocol

jasoncrawford:
## Summary

- Removes `standby` from `ForemanMessage` entirely — it had no functional effect on workers (they already wait for `task_assigned` by default)
- The foreman was sending standby on every `reconcile()` call, which fires on every GitHub webhook, causing log spam like the 4× repeated messages seen in production
- Removes all sending logic, the display entry, and ~50 lines of test code that was checking for standby responses

## Context

A proper idle acknowledgment will be part of issue #312 when the full ack/handshake lifecycle is implemented. This message is a placeholder for that, not a premature version of it.

## Test plan
- [ ] `npm test` passes (only pre-existing `foreman.http.test.ts` 404 failure)
- [ ] Confirm workers no longer log repeated "Standby: waiting for tasks..." messages in production

### DIFF
--- a/tests/foreman.admin-broadcast.test.ts
+++ b/tests/foreman.admin-broadcast.test.ts
@@ -13,6 +13,7 @@ import { TaskQueue, WorkerRegistry, createForemanWss } from "../src/foreman.js";
 import { loadDefaultConfig } from "../src/config.js";
 const defaultCfg = await loadDefaultConfig();
 import type { AdminWss, LogEntry } from "../src/admin-ws.js";
+import { waitUntil } from "./helpers.js";
 
 // ── Mock AdminWss ─────────────────────────────────────────────────────────────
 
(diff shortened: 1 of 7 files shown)


src/foreman.ts
```diff
@@ -855,20 +831,16 @@ export function createForemanWss(
...
-            sendMsg(workerId, evtMsg);
+            registry.send(workerId, { type: "event_notification", taskId: msg.taskId, event: evt });
             log(workerId, `→ event_notification #${existing.issueNumber} ${evt.name} (queued)`);
           }
         } else if (!existing) {
           log(workerId, `hello busy task=#${msg.taskId} — unknown task, respecting busy status`);
           registry.register(workerId, ws, "busy", msg.taskId);
         } else {
-          // Task is assigned to a different worker — standby
+          // Task is assigned to a different worker — register idle; assignIdleWorkers() will send standby
```
jasoncrawford: but it doesn't anymore, right?

jasoncrawford (commented):
I notice there are still references to standby in the db test, the foreman registry test, and in the smoke test—actually I wonder if the smoke test is going to fail now.

PR #406, opened 2026-03-30
feat: make DB the single source of truth for task state (issue #374)

### LINKED ISSUE #374: Refactor: make DB the single source of truth for task state
## Problem

The current architecture has two sources of truth for task state that can drift out of sync:

1. **In-memory `TaskQueue`** — the live operational state, including dependency graph (`graph`/`openIssues`), which drives all assignment logic
2. **DB `tasks` table** — a state store written to as a side effect of in-memory changes, used by the `/api/tasks` REST endpoint and the `TaskList` page

This causes several gaps:

- **Dependency/blocked state is invisible to the DB.** The `tasks` table has no `blocked` status and no knowledge of blockers, so the `TaskList` page can't distinguish a task that's pending-and-ready from one that's pending-and-blocked.
- **Status can be stale in the DB.** If a worker disconnects mid-task, the in-memory queue reverts the task to `pending`, but the DB still shows `assigned` until the task is eventually completed.
- **Recovery complexity.** The `task_assignments` table exists solely because the in-memory state doesn't survive foreman restarts. It's a third representation of the same state.
- **Two UI views show different data.** The Dashboard reads from the live WebSocket (in-memory); the TaskList reads from the DB. They can show different statuses for the same task.

## Proposed solution

Make the DB the single authoritative source of truth, and have in-memory state derive from / stay consistent with it:

- **Add `blocked` as a first-class status** in the `tasks` table. When the foreman resolves dependencies and finds open blockers, write `blocked` to the DB (not just set an in-memory flag). When a blocker closes, update to `pending`.
- **Add a `task_blockers` table** (`task_id`, `blocker_issue_number`, `closed_at`). Written once when dependencies are fetched; updated when GitHub sends a closed event.
- **On foreman restart, read from DB** instead of re-fetching all `brunel:ready` issues from GitHub. GitHub becomes a fallback for reconciliation, not the primary source.
- **Drop `task_assignments`.** Assigned tasks survive restarts naturally once the DB is the source of truth — just load `status = 'assigned'` tasks on startup.
- **`TaskList` and `Dashboard` show the same data** — both driven by DB state, with the WebSocket snapshot as the live delivery mechanism.

## Benefits

- One place to read current state
- No sync problem — writes go to DB first, in-memory reflects DB
- Blocker state visible everywhere (Dashboard, TaskList, TaskDetail)
- Simpler recovery on restart
- No more `task_assignments` table

## Notes

The main discipline required: every state change must write to DB before (or atomically with) taking effect in memory. The current code already does this for assignment (with an explicit `revertTask` on DB write failure), so the pattern exists — it just needs to be applied consistently to all transitions including `blocked`/`pending` dependency state.

jasoncrawford:
## Summary

- Adds `task_blockers` table with composite PK `(task_id, blocker_issue_number)` and `closed_at` timestamptz for tracking individual blocker relationships
- Adds `blocked` as a first-class task status in both the `tasks` DB table and in-memory `TaskQueue` — eliminates the dual source of truth where `status` lived in DB but blocking was computed separately
- Adds `TaskBlockerStore` interface/implementation (`upsertBlockers`, `closeBlocker`, `listTaskBlockers`, `listAllOpenBlockers`) in `db.ts`
- Foreman now writes `blocked`/`pending` to DB as dependencies are checked and resolved during `startDepsLoad` and issue-close events
- On foreman startup, active tasks are restored from `taskStore.listTasks()` (not `task_assignments`) — blocked and assigned tasks are reconstructed from DB state
- Removes all `assignStore.upsertAssignment`, `assignStore.deleteAssignment`, and `assignStore.updatePr` calls; `taskStore.markAssigned` is now the synchronous DB barrier in `tryAssignWork`
- Dashboard shows `blocked` count directly from `t.status`; TaskList adds a `blocked` filter button

## Test plan

- [ ] `npx tsc --noEmit` passes (verified locally)
- [ ] `npm run lint` passes (verified locally)
- [ ] `npm test` — unit + DB integration tests — verified by CI (requires Supabase)
- [ ] `npm run test:browser` — Playwright tests — verified by CI

Closes #374

### DIFF
--- a/frontend/src/types.ts
+++ b/frontend/src/types.ts
@@ -1,4 +1,5 @@
-export type { TaskStatus } from "../../shared/types.ts";
+import type { TaskStatus } from "../../shared/types.ts";
+export type { TaskStatus };
 
 export interface BlockerInfo {
   issueNumber: number;


jasoncrawford:
Good questions — two separate things here:

**On `task_blockers`**: You're right that the dependency relationships are derivable from GitHub (the issue body references). Looking at the startup sequence, Step 2 (reconstruct graph from `task_blockers`) is actually superseded by Step 3 (`loadIssuesToQueue`), which overwrites the same graph entries from GitHub via `setBlockers`. So the table isn't providing operational value it doesn't already get from GitHub.

The spec called for it to make blocked relationships visible in the DB (e.g., for a future "show why this task is blocked" UI), but the `tasks.status = 'blocked'` field already captures what's needed for the current TaskList. Happy to drop `task_blockers` entirely if you'd prefer — that would simplify the schema and the startup sequence. Let me know.

**On the `tasks` table**: It was pre-existing. What this PR added: `blocked` as a first-class status on `tasks`, the `task_blockers` table, and removal of all `task_assignments` writes (making `tasks` the single source of truth for assignment state).

supabase/migrations/20260330000000_add_blocked_status.sql
jasoncrawford: We don't need both this and the drop-table migration in the same PR! This add-table hasn't been merged yet so we can just drop it.

jasoncrawford (commented):
I notice we're still marking issues as blocked in the db. Again, isn't that derived state? Does it make sense to store it in the db? If so, are we sure that we're keeping it properly in sync?

PR #408, opened 2026-03-30
feat: add persistent worker status bar

### LINKED ISSUE #407: Worker status bar
The worker should have a status bar line at the bottom, below the prompt or below the "Working..."

It should show:

Left side, left-justified, separated by " • ":
* worker ID
* status: busy or idle
* current task (issue) # if any
* current PR # if any
* current git branch

Right side, right-justified:
* Connection status to foreman: connected, disconnected, reconnecting... (during handshake)

Example:

```
worker 7c254628 • busy • task #374 • PR #406 • db-single-source-of-truth                                Connected
```

Or:

```
worker 7c254628 • idle                                                                               Disconnected
```

Then we wouldn't need the "Connected to foreman" and "Disconnected from foreman" messages anymore, those can be dropped.

jasoncrawford:
## Summary

- Adds a persistent status bar at the bottom of the worker terminal showing worker state at all times
- Left side (bullet-separated): `worker {id8} • busy/idle • task #{N} • PR #{N} • {branch}`
- Right side: `Connected` / `Disconnected` / `Reconnecting...`
- Status bar coexists with the query "Working…" line when a query is running
- Removes the old "Connected to foreman" / "Disconnected from foreman" print messages

## Implementation details

**`src/display.ts`:**
- Extended status bar to support two layers (primary query status + persistent worker status)
- Added `startPersistentStatus` / `stopPersistentStatus` / `updatePersistentStatus`
- Added `fmtWorkerStatus()` formatter that fills terminal width with left/right content
- `_clearStatus` / `_drawStatus` now handle N active layers dynamically

**`src/worker.ts`:**
- `WorkerSession` tracks `connectionStatus`, `currentPrNumber`, `currentBranch`
- `getStatusText()` public method formats current state (also used in tests)
- `refreshBranch()` reads git branch asynchronously at start, task-assigned, query-complete
- PR number extracted from `pull_request` event notifications
- `WorkerDisplay` interface extended with optional status bar callbacks

**Tests:**
- `tests/display.worker-status.test.ts`: new tests for `fmtWorkerStatus` formatting and persistent status bar behavior
- `tests/worker.test.ts`: updated "close diagnostics" tests to check `getStatusText()` instead of print calls; added status bar content tests; updated display mock
- `tests/smoke.ts`: updated to detect worker connection via status bar "Connected" text

## Test plan

- [ ] `npm test` passes with DB
- [ ] `npx tsc --noEmit` — TypeScript clean ✓
- [ ] `npm run lint` — ESLint clean ✓
- [ ] Worker terminal shows status bar with correct content in each state

Closes #407

### DIFF
--- a/tests/repl.multiline.test.ts
+++ b/tests/repl.multiline.test.ts
@@ -74,8 +74,16 @@ describe("ask() - multiline rendering", () => {
       stdin.push("12345678"); // fills row 0 (pending-wrap at cursor=8, row 0)
       writeSpy.mockClear();   // clear setup output
       stdin.push("9");        // prevRow should be 0 → no cursor-up
-      // Must NOT emit \x1b[1A: cursor was on row 0 (pending-wrap), not row 1
-      expect(collectOutput(writeSpy)).not.toContain("\x1b[1A");
+      // Must NOT emit \x1b[1A *before* the prompt text: that would mean screenPosOf(8)
+      // incorrectly returned row=1, causing a premature cursor-up above the prompt.
+      // (A \x1b[1A *after* the prompt is fine — it's the "go back to row 0" navigation.)
+      const output = collectOutput(writeSpy);
+      const promptPos = output.indexOf("> ");
+      const firstCursorUp = output.indexOf("\x1b[1A");
+      expect(promptPos).toBeGreaterThan(-1);
+      if (firstCursorUp !== -1) {
+        expect(firstCursorUp).toBeGreaterThan(promptPos);
+      }
       stdin.push("\r");
       expect(await p).toBe("123456789");
     });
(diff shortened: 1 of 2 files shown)


jasoncrawford:
I'm also seeing some issues with the user input prompt:

* When a loop is done and the worker returns to the user input prompt, it is not displayed
* If I then hit enter, the prompt is displayed, but underneath the status bar, pushing it up

The intent is for there to be a status bar at the bottom at all times, but never for it to be pushed up the terminal.

jasoncrawford (commented):
Good point — simplified to push each part independently. `prNumber`/`branch` could theoretically arrive before a task is known (or come from a reconnecting worker), so the flat structure is also more correct.

PR #428, opened 2026-04-01
fix: forward events to worker reconnecting busy when task was unlabeled

### LINKED ISSUE #425: Honor worker task on hello-busy even if task is not brunel:ready anymore
Repro:

1. Mark a task brunel:ready
2. Have a worker start on it
3. Disconnect
4. Remove brunel:ready label
5. Reconnect worker

Observed: worker reconnects, says hello, status busy. Foreman doesn't find the task and doesn't assign it to the worker. Worker continues working on the task, but is no longer getting the notifications it needs.

Desired: foreman should give the worker the task anyway, and forward notifications.

The task doesn't even have to appear in the task list. Just make sure to forward the webhooks.

jasoncrawford:
## Summary

- When a worker reconnects claiming a busy task that is no longer in the queue (e.g., the `brunel:ready` label was removed while the worker was disconnected), the foreman now re-adds a minimal task entry to the queue so GitHub events can still be routed/forwarded to that worker.
- Previously the worker was acknowledged as busy but received no further event notifications, because `forwardEvent` routes via the task queue and the task wasn't there.
- The fix only applies when the task ID is a valid issue number (all real task IDs are); a truly invalid/non-numeric task ID still gets the old behavior (acknowledged busy, no queue entry).

## Test plan

- [ ] New test: `worker reconnecting busy with unlabeled taskId still receives event notifications for that issue` — verifies that a `routeEvent` call for the claimed issue number is forwarded as `event_notification` to the reconnected worker.
- [ ] All existing websocket tests continue to pass (including the "unknown task" tests that use the non-numeric `"nonexistent"` task ID).
- [ ] TypeScript type check passes (`npx tsc --noEmit`).
- [ ] ESLint passes (`npm run lint`).

Closes #425

### DIFF
--- a/src/foreman.ts
+++ b/src/foreman.ts
@@ -936,14 +936,14 @@ export function createForemanWss(
       // Cancel any pending reclaim timer — worker has reconnected.
       registry.cancelReclaimTimer(workerId);
 
-      if (msg.status === "busy" && msg.taskId) {
-        function flushQueuedEvents(issueRef: string | number) {
-          for (const evt of taskQueue.drainEvents(msg.taskId!)) {
-            sendMsg(workerId, { type: "event_notification", taskId: msg.taskId!, event: evt });
-            log(workerId, `→ event_notification #${issueRef} ${evt.name} (queued)`);
-          }
+      function flushQueuedEvents(taskId: string, issueRef: string | number) {
+        for (const evt of taskQueue.drainEvents(taskId)) {
+          sendMsg(workerId, { type: "event_notification", taskId, event: evt });
+          log(workerId, `→ event_notification #${issueRef} ${evt.name} (queued)`);
         }
+      }
 
+      if (msg.status === "busy" && msg.taskId) {
         const existing = taskQueue.get(msg.taskId);
         if (existing && existing.status !== "complete" && (existing.status !== "assigned" || existing.assignedWorkerId === workerId)) {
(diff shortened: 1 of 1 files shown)


jasoncrawford:
Good question — I reviewed the #406 changes carefully. The fix is still correct and no changes needed.

**Why the `!existing` case still occurs after #406:**

#406 made the DB the source of truth for task state and loads all active tasks on startup (including `assigned` ones). That covers the foreman-restart scenario. But the in-memory `!existing` case can still happen within a running foreman session:

1. Worker assigned task → task is `assigned` in queue and DB
2. Worker disconnects → reclaim timer starts
3. Label removed → webhook arrives → `labeledIssues.delete`, `reconcile()` called — task stays (it's `assigned`, not `pending`)
4. Reclaim timer fires → task reverted to `pending` in queue (and DB via `taskStore.markPending`)
5. Another webhook arrives → `reconcile()` → pending tasks not in `labeledIssues` are removed from queue
6. Worker reconnects → task not in queue → `!existing` branch → fix applies

The DB still has the task record, so when the worker eventually calls `task_complete`, `taskStore.markComplete` updates it correctly. My fix only needs to restore the in-memory queue entry (not create a new DB record) so event routing works.

I also verified: the reclaim case itself doesn't call `taskStore.markAssigned` on reconnect — it just updates in-memory state — so my fix is consistent with that pattern.

PR #453, opened 2026-04-01
fix: reset workspace and print message on hello_ack cancelled

### LINKED ISSUE #449: Worker: handle task cancellation gracefully (workspace reset + user message)
When a worker receives `hello_ack { status: "cancelled" }`, it clears task state but does two things it shouldn't:

1. **No workspace reset** — the worker leaves the workspace as-is (potentially with uncommitted changes, a stale branch, etc.). The next task it picks up starts in a dirty state. The normal task-complete path runs `workspace.reset()` via `afterTask`; cancellation needs to do the same.

2. **No user-visible message** — the task silently disappears. The user has no indication that their worker was cancelled and is now idle.

## Fix

In the `hello_ack { status: "cancelled" }` handler in `worker.ts`:

- Call `workspace.reset()` (same reset as `afterTask`) to return the workspace to a clean state
- Print a message like `"Task cancelled (reassigned to another worker). Workspace reset."` via `display.print`

The abort of the in-flight Claude session is tracked separately in #375.

jasoncrawford:
## Summary

- When a worker receives `hello_ack { status: "cancelled" }`, it now calls `workspace.reset()` to return the workspace to a clean state (same as the normal `afterTask` path)
- Prints `"Task cancelled (reassigned to another worker). Workspace reset."` via `display.print` so the cancellation is visible to the user
- If workspace reset fails, prints an error message instead

## Tests

- Added test: `prints cancellation message on hello_ack cancelled`
- Added test: `calls workspace.reset() on hello_ack cancelled when workspaceCtx is set`

Closes #449

### DIFF
--- a/tests/worker.test.ts
+++ b/tests/worker.test.ts
@@ -704,7 +704,7 @@ describe("hello_ack handshake — buffering", () => {
 
       const printed = display.print.mock.calls.map(args => stripAnsi(String(args[0]))).join("\n");
       expect(printed).toContain("cancelled");
-      expect(printed).toContain("Workspace reset");
+      expect(printed).not.toContain("Workspace reset");
     } finally {
       vi.restoreAllMocks();
       vi.useRealTimers();


src/worker.ts
```diff
@@ -396,6 +396,16 @@ export class WorkerSession {
...
         this.refreshStatus();
+        const ctx = this.options.workspaceCtx;
+        if (ctx) {
+          void ctx.workspace.reset().then(() => {
+            this.display.print(display.c.amber("Task cancelled (reassigned to another worker). Workspace reset."));
+          }).catch((err: unknown) => {
+            this.display.print(display.c.boldRed(`Task cancelled. Workspace reset failed: ${err instanceof Error ? err.message : String(err)}`));
+          });
+        } else {
+          this.display.print(display.c.amber("Task cancelled (reassigned to another worker). Workspace reset."));
```
jasoncrawford: This message says workspace reset but it's the path that doesn't reset the workspace. Maybe eliminate duplication and just move the task-canceled message above, and then have workspace-reset message be a separate thing

PR #490, opened 2026-04-02
fix: remove Supabase Docker image caching

### LINKED ISSUE #471: Supabase images not actually cached?
#405 was supposed to cache Supabase docker images so that test setup was faster. Now in a recent test run I see:

Cache Supabase Docker images: 48s
Load cached Supabase images: 2m 27s
Start local Supabase: 38s

https://github.com/jasoncrawford/brunel/actions/runs/23865681180/job/69584033801

I think that's actually slower overall then before. Why is loading supposedly-cached images still taking 2.5 minutes?

jasoncrawford:
## Summary

- The Docker image cache was making CI **slower**, not faster
- Observed timings on a cache-hit run: 48s (cache restore) + 2m27s (`docker load`) + 38s (`supabase start`) ≈ **3m53s total**
- `supabase start` pulling fresh images directly from Docker Hub takes ~2-3 minutes — comparable or better
- Root cause: `docker load` on a multi-GB tar is inherently slow, and the save step captured all Docker images (not just Supabase ones), making the archive unnecessarily large
- Also removed the "Free disk space" step that was only needed to make room for the cache tar file

## Test plan

- [ ] CI runs on this PR — verify the `test` job completes within the 15-minute timeout
- [ ] Check that "Start local Supabase" step timing is reasonable (expect ~2-3 minutes)
- [ ] No disk-space failures

Closes #471

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,13 +12,6 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Free disk space
-        run: |
-          sudo rm -rf /usr/local/lib/android
-          sudo rm -rf /usr/share/dotnet
-          sudo rm -rf /opt/ghc
-          sudo rm -rf /usr/local/.ghcup
-
       - uses: actions/setup-node@v4
         with:
           node-version: 20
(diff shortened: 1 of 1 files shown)


jasoncrawford:
Good question — here's the detailed step timing from this PR's CI run:

| Step | Time |
|---|---|
| `supabase start` (pulling images fresh) | 2m 6s |
| `npm ci` | 7s |
| `npm test` | 7s |
| `npm run coverage` | 10s |
| `tsc` + `lint` | 8s |
| **Total test job** | **~2m 45s** |

Compare to the old approach on a cache-hit run:
- Cache download: 48s
- `docker load`: 2m 27s  
- `supabase start` (images already loaded): 38s
- **Total just for Supabase setup: ~3m 53s**

So removing the cache actually sped things up by ~1m 10s overall.

The fundamental problem with the tar approach is that `docker load` on a multi-GB archive is inherently slow — it has to decompress and import every layer. And the old save step was capturing *all* Docker images on the runner (not just Supabase ones), making the archive unnecessarily large.

The real bottleneck now is the 2m 6s `supabase start`, which is the time to pull ~10 Docker images from Docker Hub. Options to improve this further:

- **Selective image cache with compression** — save only Supabase images and gzip the tar. Might shave 20-30s on cache-hit runs but adds complexity and is worse on cache-miss runs. Probably not worth it.
- **Self-hosted runner** — Docker layers would persist between runs, making `supabase start` consistently fast (~10-20s). But adds infrastructure overhead.
- **GitHub's hosted runner Docker layer cache** — GitHub doesn't expose this; Docker Hub pulls start fresh each run.

My recommendation: leave it as-is. 2m 45s for the full test job (including DB tests) is a reasonable result, and adding back a Docker cache would likely make things slower again rather than faster.

PR #505, opened 2026-04-03
fix: add worker-side heartbeat to detect stale connections

### LINKED ISSUE #485: Workers think they are connected, foreman does not
For some reason, maybe because of laptop closing and/or network loss, I have five workers in terminal windows right now who all say they are Connected in the status bar. However, the foreman shows 0 workers connected and has reset all their tasks to pending.

Turning off the wi-fi on my laptop does not immediately change the worker status bar from Connected.

Maybe the foreman is sending heartbeats, but the worker is not, and doesn't do anything if it stops receiving them?

---

Update: after opening laptop, similar thing. Here's what I see on many worker terminals:

```

WebSocket error: Client network socket disconnected before secure TLS connection was established
[worker] > 

worker elihu-91126c50 ∙ task #438 ∙ PR #481 ∙ docs-deleteTask-guard                                                                                                                                                                Connected
```

Why does the worker say connected even after it got a websocket error?

jasoncrawford:
## Summary

- Workers showed "Connected" after network drops because silent TCP disconnections don't reliably fire WebSocket `close`/`error` events
- Added a client-side ping/pong heartbeat: the worker sends a `ping` to the foreman every 20 seconds; if no `pong` comes back before the next ping, the socket is terminated, triggering the existing `close` handler and reconnect logic
- The foreman already pings workers (every 25s) but nothing on the worker side was checking liveness

## Test plan

- [x] `sends a ping after the interval when the connection is open`
- [x] `terminates connection when no pong is received before the next ping`
- [x] `does not terminate when a pong is received between pings`
- [x] `reconnects after heartbeat-induced termination`
- [x] `shows Disconnected in status bar after heartbeat timeout`
- [x] `does not ping after the connection is closed`
- [x] All 1225 non-DB tests pass

Closes #485

### DIFF
--- a/src/worker.ts
+++ b/src/worker.ts
@@ -174,6 +174,8 @@ export type WorkspaceCtx = {
 export type WorkerSessionOptions = {
   afterTask?: () => Promise<void>;
   workspaceCtx?: WorkspaceCtx;
+  /** Interval in ms between WebSocket pings. Defaults to 20_000. */
+  pingIntervalMs?: number;
 };
 
 // Sentinels used to signal WebSocket events through ask()'s abort param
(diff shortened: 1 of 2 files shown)


jasoncrawford:
This is a duplicate PR, already being worked on in #498

PR #528, opened 2026-04-05
refactor: split task models into own file

### LINKED ISSUE #500: Split out task models into own file
There is too much code in the foreman still using taskStore and taskQueue directly. And foreman.ts file is getting long anyway.

Split out all the task models into their own separate source file. That module should export TaskModel *only*, and everything the foreman server does should go through that interafce. TaskStore, TaskQueue, labeledIssues, openIssues, etc. should all be private within the taskModel module.

In general foreman should be the controller in a Model-View-Controller paradigm.

jasoncrawford:
## Summary

- Moves `TaskQueue` and `TaskModel` from `foreman.ts` into a new `src/task-model.ts` module
- Makes `TaskQueue` an internal implementation detail — only `TaskModel` is exported
- All foreman code now accesses task state exclusively through `TaskModel` methods, following an MVC pattern where `foreman.ts` acts as the controller
- `TaskModel` creates its own internal `TaskQueue` and extends `EventEmitter` to forward change events
- `createForemanWss` takes `TaskModel` instead of `TaskQueue`; `labeledIssues`, `openIssues`, and `taskStore` are now owned by `TaskModel`
- Updates all 10 foreman test files and the browser test server to use the new API

Closes #500

## Test plan

- [x] `npx tsc --noEmit` passes
- [x] `npm run lint` passes
- [x] All 1293 unit tests pass (3 DB test failures are pre-existing — require Supabase)
- [x] CI should pass all checks

### DIFF
--- a/src/foreman.ts
+++ b/src/foreman.ts
@@ -12,7 +12,7 @@ import { loadConfig } from "./config.js";
 import { setBlockers, fetchBlockers } from "./dependencies.js";
 import { fetchIssueStates } from "./github.js";
 import type { DependencyGraph } from "./dependencies.js";
-import { type DbLogger, type TaskStore, createDbLogger, createTaskStore, createNullDbLogger, createNullTaskStore, buildMessageSummary } from "./db.js";
+import { type DbLogger, createDbLogger, createNullDbLogger, buildMessageSummary } from "./db.js";
 import type { AdminWss, WorkerSnapshot } from "./admin-ws.js";
 import { fmtError } from "./utils.js";
 import { shortWorkerId } from "../shared/utils.js";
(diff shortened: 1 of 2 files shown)


src/foreman.ts
```diff
@@ -124,369 +126,8 @@ export class WorkerRegistry extends EventEmitter {
...
-      await this.store.markAssigned(taskId, workerId);
-      return true;
-    } catch {
-      this.queue.revertTask(taskId);
-      return false;
-    }
-  }
-}
+// TaskModel and Task are re-exported from src/task-model.ts for backwards compatibility.
+export { TaskModel, type Task } from "./task-model.js";
```
jasoncrawford: Why? Just fix the imports. I don't think modules should be re-exporting stuff in general, and certainly not to preserve old imports. That just obscures where things actually live and what the real dependencies are.

src/foreman.ts
```diff
@@ -9,13 +9,15 @@ import { EventEmitter } from "events";
 import type { WorkerMessage, ForemanMessage, GitHubEvent, TaskIssue, LabeledIssueState, TaskStatus } from "./types.js";
 import { fmtTimestamp, fmtEvent, setVerbose } from "./display.js";
 import { loadConfig } from "./config.js";
-import { isBlocked, setBlockers, fetchBlockers } from "./dependencies.js";
+import { setBlockers, fetchBlockers } from "./dependencies.js";
 import { fetchIssueStates } from "./github.js";
 import type { DependencyGraph } from "./dependencies.js";
 import { type DbLogger, type TaskStore, createDbLogger, createTaskStore, createNullDbLogger, createNullTaskStore, buildMessageSummary } from "./db.js";
```
jasoncrawford: I would feel better about this if the foreman did not even import the TaskStore or anything called "*taskstore*". I feel it will set up future agents to do the right thing

PR #550, opened 2026-04-05
fix: allow workers to reclaim complete tasks for finalization

### LINKED ISSUE #545: Worker knocked off of task during finalization
Problem sequence:

1. Worker completes a task
2. Worker receives "PR merged" update and is in the middle of finalization (doc updates, etc.)
3. At the same time, the PR is getting deployed to prod, which causes a server restart
4. The server restart interrupts the connection and the client reconnects
5. The client connects with hello busy
6. The foreman sees that the task is closed, and responds to tell the worker to cancel the task.
7. The client abandons everything it's doing and resets its workspace.

When a worker connects to reclaim a task, we should let it do so even if the task is complete. The only time we should cancel a task is when it has been transferred to another worker. This will allow workers to do finalization and cleanup work after an issue is closed.

jasoncrawford:
## Summary
When a worker reconnects after a server restart while doing finalization work, the task might already be marked complete (due to the issue being closed). Previously, the foreman would cancel the worker, interrupting finalization. Now, the worker is allowed to reclaim the task to finish finalization work (like doc updates, cleanup, etc.).

The only time a worker is cancelled is when the task has been transferred to another worker.

## Changes
- Remove check that cancels workers when reconnecting to complete tasks
- Only cancel workers when their task has been transferred to another worker  
- Skip `taskModel.assign()` for already-complete tasks to preserve their status
- Update tests to reflect new behavior (workers can reclaim complete tasks)
- Fix pre-existing test bug in task ordering (tasks are returned newest-first)

## Test plan
- All existing tests pass
- New test added: "cancels worker when task is assigned to a different worker"
- Updated tests: allow workers to reclaim complete tasks instead of being cancelled
- Fixed test: task ordering now matches implementation (newest-first)

Closes #545

### DIFF
--- a/src/foreman/wss.ts
+++ b/src/foreman/wss.ts
@@ -236,8 +236,7 @@ export function createForemanWss(
 
         if (!existing) {
           log(workerId, `hello busy task=#${msg.taskId} — task not found`);
-          registry.register(workerId, ws, "idle");
-          sendMsg(workerId, { type: "hello_ack", workerId, status: "cancelled" }, msg.taskId);
+          cancelWorker(msg.taskId);
         } else if (existing.assignedWorkerId && existing.assignedWorkerId !== workerId) {
           log(workerId, `hello busy task=#${msg.taskId} — task taken by another worker`);
           cancelWorker(msg.taskId);


src/foreman/wss.ts
```diff
@@ -222,9 +222,16 @@ export function createForemanWss(
...
         registry.register(workerId, ws, "busy", taskId);
-        await taskModel.assign(taskId, workerId);
+        // Only call assign if task is not already complete (to preserve task status)
+        if (!task || task.status !== "complete") {
+          await taskModel.assign(taskId, workerId);
+        } else {
+          // For complete tasks, just update the worker_id without changing status
+          // This is done implicitly by registering the worker as busy above
+          // The task stays complete while worker finishes cleanup/finalization work
+        }
```
jasoncrawford: what does taskModel.assign actually do? doesn't it set the worker on the task? shouldn't the check here be, we assign if the task is not already assigned to the worker? and then you don't need an else block because it's obvious what's going on

PR #599, opened 2026-04-10
refactor: shared db-client and Task.events static EventEmitter

### LINKED ISSUE #588: Refactor Task to active record pattern; simplify TaskModel → TaskManager
Blocked by #587

## Problem

`TaskModel` is misnamed — it's a manager/coordinator, not a model. `Task` is a read-only interface with no behavior. The real DB layer (`TaskStore`) is a hand-rolled abstraction with two implementations (Supabase + in-memory, ~220 lines of duplicated logic) that exists primarily to enable tests to run without Supabase. This was previously discussed in #557.

### 1. New `src/foreman/models/task.ts` — Task active record class

Replace `TaskStore` + `Task interface` + `TaskRow` + `rowToTask` + `deriveStatus` with a single `Task` class that talks directly to the typed Supabase client (from #587):

```typescript
let db: SupabaseClient<Database>;
let onChange: (() => void) | undefined;

export function initTask(supabase: SupabaseClient<Database>, onTaskChanged?: () => void): void {
  db = supabase;
  onChange = onTaskChanged;
}

type TaskRow = Database["public"]["Tables"]["tasks"]["Row"];

export class Task {
  // All fields as properties (typed from TaskRow)
  readonly taskId: string;
  workerId: string | null;
  // etc.

  private constructor(row: TaskRow) { /* ... */ }

  // Derived state — no DB call
  get status(): TaskStatus { /* deriveStatus logic inline */ }
  get repoUrl(): string { ... }

  // Test factory — creates an instance without hitting DB
  static fromTest(fields: Partial<TaskRow> & { task_id: string; issue_number: number }): Task

  // Static finders
  static async get(taskId: string): Promise<Task | null>
  static async getByIssue(issueNumber: number): Promise<Task | null>
  static async getByPr(prNumber: number): Promise<Task | null>
  static async getByWorker(workerId: string): Promise<Task | null>
  static async list(opts?: { cancelable?: boolean; limit?: number }): Promise<Task[]>
  static async upsert(...): Promise<Task>

  // Instance mutations — write to DB, update this in memory, then call onChange()
  async assign(workerId: string): Promise<void>
  async complete(): Promise<void>
  async revert(): Promise<void>
  async close(): Promise<void>
  async reopen(): Promise<void>
  async registerPr(prNumber, branch): Promise<void>
  async unregisterPr(): Promise<void>
  async mergePr(): Promise<void>
  async updateContent(title, body, labels): Promise<void>
  async delete(): Promise<void>
}
```

Every mutation ends with `onChange?.()`. The callback is wired at startup:
```typescript
initTask(supabase, () => taskManager.emit("changed"));
```

Callers just call `task.complete()` — the admin dashboard notification is automatic, with no caller responsibility. Mutations also update `this` fields in memory so callers see consistent state without re-fetching — this directly fixes the `Task`/`TaskRow` field divergence bug class noted in #557.

### 2. Slim `src/foreman/models/task-model.ts` → `TaskManager`

**Keep:** event queues, branch mapping, `_labeledIssues`/`_openIssues`, all issue-lifecycle methods (`isTracked`, `trackIssue`, `closeIssue`, etc.), `emit("changed")`, startup methods.

**Remove:** `store: TaskStore`, all CRUD-delegating wrappers (`get`, `getTaskForIssue`, `complete`, `revert`, `register`, `cancel`, `assign`, `registerPr`, etc.), `TaskModel.create()`.

**Rename:** `TaskModel` → `TaskManager`.

### 3. Clean up `db.ts`

Remove: `TaskRow`, `TaskStore`, `ListTasksOpts`, `createTaskStore`, `createMemoryTaskStore`.

`db.ts` becomes a pure logging module (`DbLogger`, `buildMessageSummary`, `createDbLogger`, `createNullDbLogger`).

### 4. Update callers

`wss.ts`, `event-router.ts`, `http-server.ts`, `index.ts` switch to calling `Task.*` directly:

```typescript
// Before
const task = await taskModel.getTaskForIssue(issueNumber);
await taskModel.complete(taskId);

// After
const task = await Task.getByIssue(issueNumber);
await task.complete();  // onChange fires automatically
```

### 5. Update tests

**Non-DB tests** (WebSocket protocol, event routing, reconcile, HTTP endpoints) — mock Task statics with `vi.spyOn` + `Task.fromTest()`:
```typescript
const task = Task.fromTest({ task_id: "t1", issue_number: 42, repo: "org/repo", title: "Fix bug" });
vi.spyOn(Task, "getByIssue").mockResolvedValue(task);
vi.spyOn(task, "assign").mockImplementation(async (workerId) => { task.workerId = workerId; });
```

**DB tests** (timestamp behavior, task lifecycle, constraints) — use real Supabase.

Delete `createMemoryTaskStore` — it's no longer needed.

jasoncrawford:
Two refinements to the active-record design from the comment on #588:

**1. Shared DB client (`src/foreman/db-client.ts`)**

Instead of `initTask(supabase)` binding the client inside `task.ts`, a new `db-client.ts` module exports the shared `SupabaseClient` and an `initDb(supabase)` initialiser. `index.ts` calls `initDb` once; `task.ts` imports `db` from there. This keeps the pattern extensible — any future model just imports `db` from the same place.

**2. Static `Task.events` EventEmitter**

`Task` no longer takes an `onChange` callback. Instead it exposes `Task.events`, a static `EventEmitter` that fires `"changed"` after every mutation. `index.ts` wires observers at startup:

```typescript
initDb(supabase);
Task.events.on("changed", () => taskManager.emit("changed"));
```

Any number of observers can subscribe independently without coupling `Task` to its consumers.

**Test updates**
- `setupInMemoryTasks` now emits on `Task.events` in addition to the optional `emitter` arg (backward compatible — no test call sites changed)
- DB tests (`db.tasks.test.ts`, `pipeline.test.ts`) call `initDb()` instead of `initTask()`
- Browser test server subscribes via `Task.events.on(...)` explicitly

Closes #588

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>


src/foreman/index.ts
```diff
@@ -47,7 +48,8 @@ if (isMain) {
   const supabase = createClient<Database>(config.supabaseUrl, config.supabaseSecretKey);
   const dbLogger: DbLogger = createDbLogger(supabase);
   const taskManager = new TaskManager();
-  initTask(supabase, () => taskManager.emit("changed"));
+  initDb(supabase);
+  Task.events.on("changed", () => taskManager.emit("changed"));
```
jasoncrawford: Why do we have this listen and re-emit? Why don't things that were listening to taskManager just listen to Task now? If we have to have this, then it should be the taskManager's job to listen to Task, not the main foreman's job to wire this up.

jasoncrawford: Good call. Moved the subscription into `TaskManager`'s constructor — it now wires itself to `Task.events` so `index.ts` doesn't need to know about the connection.

The re-emit is still needed because `wss.ts` listens to both `taskManager` and `registry` for dashboard broadcasts (`registry` fires `"changed"` on worker connect/disconnect), so `taskManager` stays as the aggregation point.

One consequence to note: since `TaskManager` subscribes to `Task.events` in its constructor and `Task.events` is static, each `new TaskManager()` in tests adds a listener. The `setupInMemoryTasks` helper previously double-emitted (once via `Task.events`, once directly on the emitter arg) which broke 5 tests — fixed by removing the direct call since `TaskManager` picks it up through `Task.events` automatically.

PR #605, opened 2026-04-11
refactor: replace hand-written TaskRow with type derived from DB schema

jasoncrawford:
## Summary

- Moves `src/database.types.ts` → `shared/database.types.ts` so both the backend and frontend can import the generated Supabase types (the frontend tsconfig includes `shared/` but not `src/`)
- Replaces the hand-written `TaskRow` interface in `frontend/src/types.ts` with a type in `shared/types.ts` whose field types are indexed from `Row<"tasks">` — so column type changes in migrations are caught at compile time without updating two separate declarations
- Updates all backend/test imports from `src/database.types.js` to `shared/database.types.js`
- Updates CLAUDE.md to reflect the new file location and regeneration command

Zero behavior change: the HTTP wire format is unchanged; only the TypeScript type definitions are restructured.

## Test plan

- [x] `npx tsc --noEmit` passes (backend)
- [x] `npm test` — all 1310 unit tests pass (3 DB-only failures are expected without a running Supabase instance)
- [x] `npm run lint` — clean

Closes #592

### DIFF
--- a/shared/types.ts
+++ b/shared/types.ts
@@ -4,7 +4,25 @@ import type { Row } from "./database.types.js";
 
 export type TaskStatus = "pending" | "assigned" | "pushed" | "merged" | "closed" | "complete" | "blocked";
 
-// The shape of a task row as returned by GET /api/tasks: the DB row plus a
-// derived status field. Using Row<"tasks"> ensures column type changes in
-// migrations are caught at compile time.
-export type TaskRow = Row<"tasks"> & { status: TaskStatus };
+// The shape of a task as returned by GET /api/tasks (camelCase for JS
+// conventions). Field types are indexed from Row<"tasks"> so that column type
+// changes in migrations are caught at compile time. Task.toJSON() is annotated
+// with this return type, so the compiler enforces that the two stay in sync.
+type DbRow = Row<"tasks">;
+export type TaskRow = {
+  taskId: DbRow["task_id"];
+  issueNumber: DbRow["issue_number"];
+  repo: DbRow["repo"];
+  title: DbRow["title"];
+  body: DbRow["body"];
+  labels: DbRow["labels"];
+  status: TaskStatus;
+  workerId: DbRow["worker_id"];
+  prNumber: DbRow["pr_number"];
+  branch: DbRow["branch"];
+  createdAt: DbRow["created_at"];
+  assignedAt: DbRow["assigned_at"];
+  completedAt: DbRow["completed_at"];
+  issueClosedAt: DbRow["issue_closed_at"];
+  prMergedAt: DbRow["pr_merged_at"];
+};
--- a/shared/types.ts
+++ b/shared/types.ts
@@ -1,3 +1,23 @@
 // Types shared between the backend (src/) and frontend (frontend/src/).
 
+import type { Database } from "./database.types.js";
+
 export type TaskStatus = "pending" | "assigned" | "pushed" | "merged" | "closed" | "complete" | "blocked";
+
+// The shape of a task row as returned by GET /api/tasks.
+// Field types are derived from the generated DB schema so that column type
+// changes in migrations are caught at compile time.
+type DbTaskRow = Database["public"]["Tables"]["tasks"]["Row"];
+export type TaskRow = {
+  taskId: DbTaskRow["task_id"];
+  issueNumber: DbTaskRow["issue_number"];
+  repo: DbTaskRow["repo"];
+  title: DbTaskRow["title"];
+  status: TaskStatus;
+  workerId: DbTaskRow["worker_id"];
+  prNumber: DbTaskRow["pr_number"];
+  branch: DbTaskRow["branch"];
+  createdAt: DbTaskRow["created_at"];
+  assignedAt: DbTaskRow["assigned_at"];
+  completedAt: DbTaskRow["completed_at"];
+};
--- a/shared/types.ts
+++ b/shared/types.ts
@@ -1,23 +1,10 @@
 // Types shared between the backend (src/) and frontend (frontend/src/).
 
-import type { Database } from "./database.types.js";
+import type { Row } from "./database.types.js";
 
 export type TaskStatus = "pending" | "assigned" | "pushed" | "merged" | "closed" | "complete" | "blocked";
 
-// The shape of a task row as returned by GET /api/tasks.
-// Field types are derived from the generated DB schema so that column type
-// changes in migrations are caught at compile time.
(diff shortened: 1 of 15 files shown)


shared/types.ts
```diff
@@ -1,3 +1,23 @@
...
 
+import type { Database } from "./database.types.js";
+
 export type TaskStatus = "pending" | "assigned" | "pushed" | "merged" | "closed" | "complete" | "blocked";
+
+// The shape of a task row as returned by GET /api/tasks.
+// Field types are derived from the generated DB schema so that column type
+// changes in migrations are caught at compile time.
+type DbTaskRow = Database["public"]["Tables"]["tasks"]["Row"];
+export type TaskRow = {
```
jasoncrawford: Why do we need both DbTaskRow and a separate TaskRow? I thought the idea was to have neither of these, and do everything with `Row<"tasks">`.

jasoncrawford (commented):
Good point — you're right. The intermediate `DbTaskRow` alias was unnecessary, and the separate hand-written `TaskRow` object type should be replaced with `Row<"tasks"> & { status: TaskStatus }` directly. To make that work cleanly I'll also update `Task.toJSON()` to return snake_case fields (matching the DB row), and update the frontend components and tests accordingly. This eliminates the camelCase/snake_case impedance mismatch between the API and the DB schema.

jasoncrawford (commented):
Oh crap, I didn't realize that having a single task type was going to lead to using snake_case everywhere. That's pretty ugly.

If that's the case then clearly `Row<"tasks">` can only be used for db access. But do we need a separate TaskRow thing with the camelCase attributes? Or could we just use Task for that somehow?

Maybe the front end needs its own set of types? It would be nice to not have to duplicate them though.

jasoncrawford:
You've hit the core tension: `Row<"tasks">` is the DB schema type which is snake_case by definition, so using it directly forces snake_case everywhere in the frontend. The two options are:

1. **Snake_case throughout** (current): `TaskRow = Row<"tasks"> & { status }` — clean type, ugly JS property names in the frontend
2. **CamelCase API type** (my first approach): define `TaskRow` in `shared/` with camelCase names, but index the field *types* from `Row<"tasks">` so type safety still flows from the schema

I think option 2 is right. The frontend shouldn't have to speak the DB's naming convention. The trick to avoid pure duplication is to annotate `Task.toJSON()` with the explicit `TaskRow` return type — that way the compiler enforces that the two definitions stay in sync. If `worker_id` changes from `string | null` to `string` in the schema, `DbRow["worker_id"]` changes, `TaskRow.workerId` changes, and `toJSON()` fails to compile until it returns the right value.

Reverting to camelCase now.

jasoncrawford (commented):
OK, I get why TaskRow and Row<"tasks"> are different. But why do we have to have TaskRow when Task exists? Feels redundant?

jasoncrawford:
Good question. `Task` can't be used in the frontend because the frontend tsconfig only includes `frontend/src/` and `shared/` — `src/` is excluded. Even if that weren't the case, `Task` is a class with a private constructor, DB-dependent instance methods, and in-memory state that has no meaning in the browser.

`TaskRow` is the DTO (the plain JSON object that crosses the HTTP boundary). `Task.toJSON(): TaskRow` is what ties them together: the compiler enforces that whenever `Task` changes its serialization, it must still satisfy `TaskRow`. So there's no duplication — there's one definition (`TaskRow`) and one place where it's verified against the class (`toJSON()`'s return type annotation).

This is a pretty standard backend-class / wire-DTO split. Happy to discuss further if you see a cleaner way.

jasoncrawford:
Why do we still have TaskRow and also TaskSnapshot?

jasoncrawford:
They travel over different channels and carry different data:

**`TaskRow`** — the full HTTP response from `GET /api/tasks` (`Task.toJSON()`). Contains all DB fields: `body`, `labels`, all timestamps (`issueClosedAt`, `prMergedAt`, etc.). Used by `TaskList` on initial load.

**`TaskSnapshot`** — a lightweight WebSocket push from `Task.toSnapshot()`, sent inside `AdminMessage` over the admin WS connection. Contains only the fields the live dashboard needs: `taskId`, `title`, `status`, `assignedWorkerId`, `prUrl` (computed), and `blockers` (in-memory state that doesn't exist in the DB at all). Used by `Dashboard`, `TaskDetail`, and by `TaskList` for live status merges after initial load.

The two types couldn't easily be merged: `TaskSnapshot` includes `blockers` and `prUrl` which aren't DB columns, uses `assignedWorkerId` rather than `workerId`, and intentionally omits the heavy fields. Neither is a subset of the other.

jasoncrawford:
You're right that there's overlap — both render task tables and both end up on the frontend. The split exists because they come from different transports: `TaskSnapshot` is what the WebSocket broadcasts (kept minimal because it's pushed on every change), and `TaskRow` is what the HTTP endpoint returns (fuller because it's fetched once on page load).

That said, the architectural question of whether to unify them or eliminate one transport path is real and worth considering — it's just out of scope for this PR, which is only about removing the hand-written interface and deriving the types from the schema. Happy to file a follow-up issue for that if you'd like.

For now, is this PR acceptable as-is, or would you prefer to block on that question?

shared/types.ts
```diff
@@ -1,3 +1,28 @@
...
+  status: TaskStatus;
+  workerId: DbRow["worker_id"];
+  prNumber: DbRow["pr_number"];
+  branch: DbRow["branch"];
+  createdAt: DbRow["created_at"];
+  assignedAt: DbRow["assigned_at"];
+  completedAt: DbRow["completed_at"];
+  issueClosedAt: DbRow["issue_closed_at"];
+  prMergedAt: DbRow["pr_merged_at"];
+};
```
jasoncrawford: Rather than hardcode this, is there some way to generate it? Either generate these at build time along with the generated supabase types, or generate them at runtime. Otherwise we're going to have to (a) keep this up to date as fields change and (b) do this again for every other db type we send over the wire.

jasoncrawford:
TypeScript 4.1+ template literal types can do the snake_case → camelCase key transformation at the type level — no build step or codegen needed. Adding `SnakeToCamel` and `CamelCaseKeys` utilities to `shared/types.ts` so `TaskRow = CamelCaseKeys<Row<"tasks">> & { status: TaskStatus }`. Any new column added to the schema will automatically appear as a camelCase key in `TaskRow`, and `Task.toJSON(): TaskRow` will fail to compile until the mapping is updated. Implementing now.

jasoncrawford:
Nevermind. I give up on this whole thing. I think it was a mistake. Let's just cancel this.

PR #616, opened 2026-04-11
feat: active-record models for WebhookEvent and ForemanMessage

jasoncrawford:
## Summary

- Add `WebhookEvent` active-record class (`webhook_events` table) with `fromIncoming()`, fire-and-forget `log()`, `toWorkerPayload()`, `format()`, and query methods
- Add `ForemanMessage` active-record class (`foreman_messages` table) with fire-and-forget `log()`, `buildSummary()` as single source of truth for log summaries, and query methods
- Add `queryActivityLog()` in `activity-log.ts` that merges both tables sorted by timestamp
- Delete `db.ts` — `DbLogger` interface and helpers are superseded by the new model classes
- Rename wire type `ForemanMessage` → `ForWorkerMsg` in `types.ts` to avoid naming collision with the new DB class
- Rename `GitHubEvent` → `WorkerEvent` in `types.ts` (the agent-side event payload type)

## Test plan

- [ ] `npm test` passes (1318 tests; DB tests require `supabase start`)
- [ ] `npm run lint` passes
- [ ] `db.test.ts` covers `WebhookEvent.log`, `ForemanMessage.log`, `ForemanMessage.buildSummary`, `queryActivityLog` with real Supabase

Closes #593


jasoncrawford (commented):
This PR is largely on the right track — active records for `WebhookEvent` and `ForemanMessage`, `toWorkerPayload()` as the serialization seam, deleting `DbLogger`, fire-and-forget logging. All good.

One thing has changed since this branch was cut: we just merged a type system design guide (`docs/type-system.md`, also summarized in CLAUDE.md under "Type system"). That doc crystallizes some patterns that affect the naming and structure here. Specific things to revisit:

---

**1. Wire types should live in `src/wire.ts`, imported as a namespace**

The design calls for all wire types in a single `src/wire.ts`, imported everywhere as:

```ts
import * as Wire from "./wire.js";
// Wire.ForemanMessage, Wire.WebhookEvent, etc.
```

Right now `ForWorkerMsg` and `WorkerEvent` live in `src/types.ts`. They should move to `src/wire.ts`.

---

**2. `ForWorkerMsg` → `Wire.ForemanMessage`**

`ForWorkerMsg` is a bad name — it was invented to avoid colliding with the new `ForemanMessage` DB class. Once wire types live in `wire.ts` and are accessed as `Wire.ForemanMessage`, there's no collision: `ForemanMessage` (the DB model class) and `Wire.ForemanMessage` (the protocol union type) are unambiguously different things. The renaming to `ForWorkerMsg` and the downstream rename of `printForemanMessage` → `printForWorkerMsg` can all be reverted.

---

**3. `WorkerEvent` → `Wire.WebhookEvent`**

`WorkerEvent` is `{ id, name, payload }` — the wire representation of a GitHub webhook event as seen by the worker. Under the design, that's `Wire.WebhookEvent`. The unadorned name belongs to the concept; the `Wire.` prefix carries the context that it's the lean wire shape.

---

**4. Rename the `ForemanMessage` DB model class**

The DB model class isn't really "a foreman message" — it's a log entry for a protocol message. A name like `MessageLog` (or `ProtocolMessage`) would be more accurate and is what frees up `ForemanMessage` for the wire type (point 2 above). This is the one rename that touches the new model file itself.

---

**5. Remove `WebhookEventData` / `ForemanMessageData` input interfaces (if present)**

The design says: no `XxxData` interfaces that mirror a model class's fields. If these exist as separate interfaces just to type constructor/factory inputs, replace them with named params or remove them — the class itself is the type.

---

The structural work here (active records, `toWorkerPayload()`, deleting `DbLogger`) is right and doesn't need to change. The above is mostly naming and the `wire.ts` consolidation.

jasoncrawford:
Thanks for the review! I've implemented all the requested changes:

1. **Created `src/wire.ts`** — all wire types (`ForemanMessage`, `WebhookEvent`, `WorkerMessage`, `TaskIssue`) consolidated here, imported as `import * as Wire from "./wire.js"` everywhere.

2. **`ForWorkerMsg` → `Wire.ForemanMessage`** — reverted the awkward rename; the `Wire.` namespace prefix provides the disambiguation from the DB model class.

3. **`WorkerEvent` → `Wire.WebhookEvent`** — consistent with the convention.

4. **`ForemanMessage` DB model → `MessageLog`** — renamed via `git mv` (preserves history); file is now `src/foreman/models/message-log.ts`. The `Wire.ForemanMessage` vs `MessageLog` naming is now unambiguous.

5. **Removed `WebhookEventData` / `ForemanMessageData`** — types are inlined directly in the `log()` method signatures; no separate exported interfaces.

`toWorkerPayload()` return type is now `Wire.WebhookEvent`. CLAUDE.md updated to reflect the new structure.

jasoncrawford:
Correction to my earlier review: point 4 (renaming the `ForemanMessage` DB class to `MessageLog`) was wrong. `ForemanMessage` is a fine entity name that matches the DB table — no rename needed. The `Wire.` namespace solves the collision on its own: `ForemanMessage` is the DB model, `Wire.ForemanMessage` is the protocol type. They're unambiguous without either changing.

jasoncrawford:
Got it — reverted. `ForemanMessage` stays as the DB model class name; `Wire.ForemanMessage` vs `ForemanMessage` is unambiguous on its own.

src/types.ts
jasoncrawford, line 1: This is a pretty useless file now, can it be removed?

PR #618, opened 2026-04-11
refactor: convert commands.ts to CommandRegistry class

### LINKED ISSUE #612: Convert commands.ts to CommandRegistry class
The module-level functions in `commands.ts` (`register`, `lookup`, `listAll`, `execute`, `_reset`, `scoped`) are effectively an object in disguise. Convert to a `CommandRegistry` class.

## Changes

- `CommandRegistry` class with all current functions as instance methods
- `scoped(prefix)` returns a `ScopedCommandRegistry` object (with its own `register` / `scoped` methods) that delegates to the parent registry rather than returning a bare function
- Registry is instantiated in `src/agent/index.ts` and injected into modules that need it (e.g. passed to `registerWorkspaceCommands`, `registerWorkerCommands`, `handleModelCommand`, `handleEffortCommand`, `dispatchInput`, etc.)
- `_reset()` stays as a test-only instance method

## Affected files

- `src/agent/command-registry.ts` — becomes the class definition (renamed from `commands.ts`)
- `src/agent/input.ts` — receives registry (or calls via injected ref)
- `src/agent/workspace.ts` — `registerWorkspaceCommands(deps, registry, workerMode?)`
- `src/agent/worker.ts` — `registerWorkerCommands(session, registry)`
- `src/agent/model.ts` — receives registry to register `/model`
- `src/agent/effort.ts` — receives registry to register `/effort`
- `src/agent/index.ts` — instantiates registry, wires it into everything
- Test files that call `_reset()` or `register()` directly

jasoncrawford:
## Summary

- Replaces the module-level global state in `commands.ts` with a `CommandRegistry` class (and `ScopedCommandRegistry` helper for namespaced registration)
- Registry is instantiated once in `index.ts` and injected into all modules that need it (`registerWorkspaceCommands`, `registerWorkerCommands`, `dispatchInput`, `listCommandNames`, etc.)
- Adds `registerModelCommand(registry, deps)` to `model.ts` and `registerEffortCommand(registry, deps)` to `effort.ts` so those modules own their own command registration
- All functions in `input.ts` that read from the registry (`parseSlashCommand`, `dispatchInput`, `listCommandNames`, `listWorkerCommandNames`, `listCommands`, `listWorkerCommands`) now accept a `CommandRegistry` as their first parameter
- `_reset()` is preserved as an instance method for test isolation
- All affected tests updated to create `CommandRegistry` instances directly

## Test plan

- [ ] `npm test` — all 1318 unit tests pass (3 skipped DB tests require Supabase)
- [ ] `npm run lint` — no lint errors

Closes #612


src/agent/commands.ts
```diff
@@ -13,58 +13,83 @@ export interface CommandEntry {
   handler: CommandHandler;
 }
 
-const _entries: Map<string, CommandEntry> = new Map();
+/**
+ * A registry scoped to a given prefix.
+ * e.g. registry.scoped("workspace").register("create", opts) registers "workspace:create".
+ */
+export class ScopedCommandRegistry {
```
jasoncrawford: Do we really need a separate class for this? if CommandRegistry had an optional namespace, then scoping could just give you back another CommandRegistry. Any client of CommandRegistry wouldn't even need to know if it was namespaced or not.

I guess it would need an optional parent too. Maybe this isn't worth it? Still, since they expose the same interface, I like the idea that a client doesn't need to know what kind they have.

src/agent/commands.ts
```diff
@@ -13,58 +13,83 @@ export interface CommandEntry {
...
- * Return all command entries available in the given mode.
- * In non-worker mode, commands with availability "worker" are excluded.
- * In worker mode, commands with availability "repl" are excluded.
- */
-export function listAll(workerMode = false): CommandEntry[] {
-  return Array.from(_entries.values()).filter(e =>
-    workerMode ? e.availability !== "repl" : e.availability !== "worker",
-  );
-}
+/** Registry of slash commands available in the REPL and/or worker mode. */
```
jasoncrawford: we have dissolved repl vs. worker mode; all commands available at all times. maybe the concept of availability can go away?

src/agent/commands.ts
jasoncrawford, line 1: let's rename this file command-registry.ts

src/agent/index.ts
```diff
@@ -371,7 +365,7 @@ export async function main(
     // empty promptLine suppresses the drawFresh callback so incoming messages
     // are printed cleanly without a prompt preceding or following them.
     const promptStr = session ? (showPrompt ? "\n[agent] > " : "") : "\n> ";
-    const input = await ask(promptStr, listWorkerCommands, wsAbort);
+    const input = await ask(promptStr, () => listWorkerCommands(registry), wsAbort);
```
jasoncrawford: Why isn't listCommands a method on registry? I would have expected this to be `() => registry.listCommands()`. Also we don't have worker vs. repl commands anymore.

… Hmm, I guess this is about walking directories to find skills and file-based commands? Question for discussion: should that responsibility also be in CommandRegistry?

src/agent/worker.ts
```diff
@@ -675,9 +675,9 @@ export class WorkerSession {
  *
  * Follows the same pattern as registerWorkspaceCommands in workspace.ts.
  */
-export function registerWorkerCommands(session: WorkerSession | undefined): void {
-  const workerReg = scoped("worker");
-  workerReg("complete", {
+export function registerWorkerCommands(session: WorkerSession | undefined, registry: CommandRegistry): void {
+  const workerReg = registry.scoped("worker");
```
jasoncrawford: I tend to think the scoping should be done in index.ts, and that should pass a scoped registry into this method

src/agent/workspace.ts
```diff
@@ -191,14 +191,14 @@ export interface WorkspaceCommandDeps {
...
  * Register workspace commands into the command registry.
- * Call this once at startup (or in tests via beforeEach after _reset()).
+ * Call this once at startup (or in tests via beforeEach after registry._reset()).
  * If workerMode is true, workspace:create prints "managed automatically" instead
  * of creating a workspace (workers have their workspace managed by the foreman).
  */
-export function registerWorkspaceCommands(deps: WorkspaceCommandDeps, workerMode = false): void {
-  const reg = scoped("workspace");
+export function registerWorkspaceCommands(deps: WorkspaceCommandDeps, registry: CommandRegistry, workerMode = false): void {
+  const reg = registry.scoped("workspace");
```
jasoncrawford: again, maybe scoping should be done at top level?

src/agent/index.ts
```diff
@@ -297,47 +297,41 @@ export async function main(
...
-      const newEffort = await handleEffortCommand(args, currentEffort, pickEffortFn, display.print);
-      currentEffort = newEffort;
-      if (session) session.currentEffort = newEffort;
-    },
+  registerEffortCommand(registry, {
+    getCurrentEffort: () => currentEffort,
+    setCurrentEffort: (e) => { currentEffort = e; if (session) session.currentEffort = e; },
+    pickFn: (opts, idx) => pick(opts, { currentIdx: idx, escapable: true }),
+    print: display.print,
   });
```
jasoncrawford, line 348: Not sure it improved things to move these registrations into the effort.ts and model.ts. We had to push a lot of deps across that module boundary to make that happen. Maybe keep them here?

src/agent/index.ts
jasoncrawford: Good question. I've left it in `input.ts` for now — the directory-walking logic (`~/.claude/commands/`, skill discovery, frontmatter parsing) feels filesystem-specific enough that folding it into a data registry would muddle the registry's responsibilities. Worth revisiting as a follow-up if we want to consolidate it.

PR #626, opened 2026-04-11
refactor: move listCommands/listCommandNames into CommandRegistry

### LINKED ISSUE #621: Consider moving listCommands into CommandRegistry
Raised during review of #618.

Currently `listCommands` / `listCommandNames` live in `input.ts` and are standalone functions that take a `CommandRegistry` as their first argument. The reviewer asked whether this should be a method on `CommandRegistry` (i.e. `registry.listCommands()`).

Rationale for keeping it in `input.ts`: the function also walks `~/.claude/commands/` for file-based commands and discovers skills from `~/.claude/skills/` and `installed_plugins.json`. This filesystem/config-layer responsibility feels out of place in a data registry.

Rationale for moving it: a caller would have a single `registry.listCommands()` call without needing to import `listCommands` separately, and `CommandRegistry` would be the single place to query "what commands are available?".

Worth revisiting as a follow-up. One option: keep `CommandRegistry` pure but add a higher-level `CommandSource` or `CommandIndex` that wraps registry + filesystem lookup.

jasoncrawford:
## Summary

- `listCommandNames` and `listCommands` are now methods on `CommandRegistry` rather than standalone functions in `input.ts` that took the registry as their first argument — callers use `registry.listCommands()` instead of `listCommands(registry)`
- Related helpers (`parseFrontmatter`, `resolveCommandFilePath`, `resolveContent`, `listSkillNames`, `ListDir`, `CommandSuggestion`) moved to `command-registry.ts` and re-exported from `input.ts` for backward compat
- Injectable `listDir`/`readFile` params preserved as method arguments, keeping testability unchanged

## Test plan

- [ ] All existing `listCommandNames` / `listCommands` tests updated to use registry methods and still pass
- [ ] `npm test` passes (1313 tests)
- [ ] `npm run lint` passes

Closes #621

### DIFF
--- a/src/agent/command-registry.ts
+++ b/src/agent/command-registry.ts
@@ -1,5 +1,7 @@
 // ── Command Registry ──────────────────────────────────────────────────────────
 
+import fs from "fs";
+
 export type HandlerResult = void | "exit" | "task-complete";
 export type CommandHandler = (args: string) => Promise<HandlerResult>;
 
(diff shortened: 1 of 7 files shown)


src/agent/command-registry.ts
jasoncrawford, line 1: This diff includes a lot of new exports. Are they all part of this class's public interface? They look more like private/internal helpers to me.

src/agent/input.ts
```diff
@@ -1,6 +1,8 @@
-import fs from "fs";
 import * as display from "./display.js";
-import type { CommandRegistry } from "./command-registry.js";
+import { defaultReadFile, resolveContent } from "./command-registry.js";
+import type { CommandRegistry, CommandSuggestion } from "./command-registry.js";
+export { parseFrontmatter, resolveCommandFilePath, resolveContent, listSkillNames, defaultReadFile } from "./command-registry.js";
+export type { ListDir, CommandSuggestion } from "./command-registry.js";
```
jasoncrawford: Are all of these imports and exports necessary? I think most if not all of them can be deleted

src/agent/index.ts
```diff
@@ -20,7 +20,7 @@ import type { ModelInfo, FetchModelsFn } from "./model.js";
 import { handleEffortCommand } from "./effort.js";
 import type { EffortValue } from "./effort.js";
 import { CommandRegistry } from "./command-registry.js";
-export { parseSlashCommand, resolveCommandFilePath, resolveContent, dispatchInput, matchCommands, listCommandNames, listCommands, ask } from "./input.js";
+export { parseSlashCommand, resolveCommandFilePath, resolveContent, dispatchInput, matchCommands, ask } from "./input.js";
```
jasoncrawford: do we need to do all of these imports? how many of these are now private/internal to CommandRegistry?

jasoncrawford (commented):
Good catches — pushed a cleanup commit (8590918):

- **`input.ts` re-exports**: removed entirely. Tests now import `parseFrontmatter`, `listSkillNames`, `resolveCommandFilePath`, `resolveContent`, `ListDir`, `CommandSuggestion` directly from `command-registry.ts`. `input.ts` still imports `resolveContent` internally (for `dispatchInput`) and has its own local `defaultReadFile`.

- **`command-registry.ts` exports**: `defaultReadFile` is now unexported (pure internal detail). The remaining new exports — `parseFrontmatter`, `resolveCommandFilePath`, `resolveContent`, `listSkillNames` — are kept exported since they each have direct tests and aren't implementation details of `CommandRegistry` itself (they're independent utility functions it delegates to).

- **`index.ts` exports**: removed `resolveCommandFilePath`, `resolveContent`, and `ListDir` — these no longer need to tunnel through `input.ts`.

PR #627, opened 2026-04-11
refactor: rename WorkerStatusModel to AgentStatus; always generate agentId at startup

### LINKED ISSUE #623: Rename WorkerStatusModel to AgentStatus; always generate agentId at startup
## Background

`WorkerStatusModel` in `src/agent/worker.ts` is the reactive model backing the persistent status bar. It already holds `model` and `effort`, but currently only exists in worker mode. `main()` in `index.ts` tracks `currentModel` and `currentEffort` as separate local vars and manually syncs them to the session — two sources of truth.

## Proposed changes

**Rename `WorkerStatusModel` → `AgentStatus`**

The model backs the status bar in both REPL and worker modes, so the name should reflect that.

**Rename `workerId` → `agentId`; always generate at startup**

Currently `workerId` is only generated inside `startWorkerMode()`. Instead, generate `agentId` unconditionally at startup (before any mode fork), and pass it into `AgentStatus`. Worker mode uses the same `agentId` as the foreman identity — no separate ID needed.

Side benefit: `sessionId_` (used to name the workspace directory) and `workerId` are currently separate UUIDs. With `agentId` always set upfront, these converge — the agent uses its `agentId` for both workspace naming and foreman handshake.

Note: `worker_id` stays as-is in the DB — when the foreman is talking to an agent, it's a worker by definition.

**Create `AgentStatus` unconditionally in `main()`**

`main()` drops the local `currentModel`/`currentEffort` vars and reads from the model instead. In worker mode, `startWorkerMode` receives the existing `AgentStatus` rather than creating its own.

## Benefit

- Single source of truth for model/effort — no manual sync between `main()` and `WorkerSession`
- Lays groundwork for making REPL and worker modes more fluid: an agent started in REPL mode already has an identity and can connect to the foreman later without restructuring

jasoncrawford:
## Summary

- Renames `WorkerStatusModel` → `AgentStatus` (the model backs the status bar in both REPL and worker modes, not just worker mode)
- Renames `workerId` constructor param → `agentId` on `AgentStatus`; types `effort` as `EffortValue` throughout
- Generates `agentId` unconditionally at startup in `main()` before any mode fork, replacing both the separate `sessionId_` UUID (REPL workspace naming) and the `workerId` generated inside `startWorkerMode()`
- Creates `AgentStatus` unconditionally in `main()`; drops local `currentModel`/`currentEffort` vars — `main()` now reads from `agentStatus.model/effort` as the single source of truth
- `WorkerSession` receives the pre-built `AgentStatus` instead of creating its own and drops its own model/effort getters/setters
- `startWorkerMode()` accepts `AgentStatus` as a second parameter instead of generating a fresh workerId internally

## Test plan

- [ ] `npm test` — all 1313 non-DB unit tests pass
- [ ] `npm run lint` — no lint errors
- [ ] `npm run smoke` — foreman + worker connect successfully

Closes #623

### DIFF
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -13,7 +13,7 @@ import { AgentStatus, WorkerSession, registerWorkerCommands, startWorkerMode } f
 import type { RunQuery, WorkerModeConfig } from "./worker.js";
 import { loadConfig } from "../config.js";
 import { Workspace, confirmIfUnsafe, registerWorkspaceCommands } from "./workspace.js";
-import { fmtError, generateWorkerId } from "../utils.js";
+import { fmtError, generateAgentId } from "../utils.js";
 import { handleModelCommand, getCachedModels, _resetCachedModels, setCachedModels } from "./model.js";
 import type { ModelInfo, FetchModelsFn } from "./model.js";
 import { handleEffortCommand } from "./effort.js";
@@ -251,9 +251,8 @@ export async function main(
   initialEffort?: EffortValue,
 ): Promise<void> {
   // Generate agentId unconditionally — used as both workspace identity and foreman identity.
-  const agentId = generateWorkerId();
-  const agentStatus = new AgentStatus(agentId);
-  agentStatus.update({ model: initialModel, effort: initialEffort });
+  const agentId = generateAgentId();
+  const agentStatus = new AgentStatus(agentId, { model: initialModel, effort: initialEffort });
 
   // Worker mode setup: create workspace, session, signal handlers.
   const workerCtx = workerConfig ? await startWorkerMode(workerConfig, agentStatus) : undefined;
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -1,5 +1,4 @@
 import "dotenv/config";
-import crypto from "node:crypto";
 import os from "node:os";
 import path from "node:path";
 import fs from "fs";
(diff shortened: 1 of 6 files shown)


src/agent/index.ts
```diff
@@ -251,18 +250,21 @@ export async function main(
   initialModel?: string,
   initialEffort?: EffortValue,
 ): Promise<void> {
+  // Generate agentId unconditionally — used as both workspace identity and foreman identity.
+  const agentId = generateWorkerId();
```
jasoncrawford: rename this to generateAgentId?

src/agent/index.ts
```diff
@@ -251,18 +250,21 @@ export async function main(
...
+  const agentStatus = new AgentStatus(agentId);
+  agentStatus.update({ model: initialModel, effort: initialEffort });
```
jasoncrawford: maybe the constructor could take initial attrs? so you don't have to immediately update

src/agent/index.ts
jasoncrawford: Done — renamed to `generateAgentId` and updated the comment too.

src/agent/index.ts
jasoncrawford: Done — added an optional `initial` parameter to the constructor (`Omit<WorkerStatusPatch, 'reconnectAt'>` since reconnectAt doesn't make sense at construction time). The call site is now `new AgentStatus(agentId, { model: initialModel, effort: initialEffort })`.

PR #633, opened 2026-04-11
feat: extract QueryStats class for per-query stat accumulation

### LINKED ISSUE #624: Extract QueryStats model for per-query transient state in runQuery
## Background

`runQuery` in `src/agent/index.ts` tracks per-query stats as a plain local object:

```ts
const stats = { turns: 0, inputTokens: 0, completedOutputTokens: 0, currentOutputTokens: 0 };
```

`startTime` is also a bare local. Together these drive the status bar text via a `getStatusText` closure that `display.startStatus` polls on a timer.

## Proposed change

Extract a small `QueryStats` class (or similar) that owns these fields and exposes an `update(event)` method for processing stream events. This object lives in `index.ts` (it's per-query, not shared).

Optionally, have `QueryStats` emit `"change"` events so the status bar can subscribe reactively rather than poll — consistent with the reactive model pattern used by `AgentStatus` / `WorkerStatusModel`.

## Benefit

- Gives a named home to the token/turn accumulation logic, which is currently scattered across three inline conditionals in the stream loop
- Makes the stats independently testable (feed it stream events, assert on the output)
- If reactive: eliminates the `getStatusText` polling pattern in favour of event subscription, consistent with the rest of the codebase

jasoncrawford:
## Summary

- Extracts a `QueryStats` class (`src/agent/query-stats.ts`) from the inline stat-tracking logic in `runQuery`
- The class owns `turns`, `inputTokens`, `completedOutputTokens`, and `currentOutputTokens`, plus `startTime`
- Exposes an `update(event)` method that processes `message_start`, `message_delta`, and `message_stop` stream events and emits `"change"` on each stat-bearing event — consistent with the `AgentStatus` reactive model pattern
- Exposes `getStatusText(workingVerb)` for status bar rendering, replacing the `getStatusText` closure in `runQuery`
- 26 new unit tests covering initial state, per-event-type behavior, multi-turn accumulation, elapsed time, and status text formatting

## Test plan

- [ ] Existing `repl.query.test.ts` tests still pass (stat accumulation behavior unchanged)

Closes #624

### DIFF
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -113,9 +113,6 @@ export async function runQuery(
     process.stdout.write("\r\n\x1b[J");
   }
 
-  // Accumulate stats from stream_event messages to show in the status line.
-  // message_delta.usage.output_tokens is cumulative per message, so we sum
-  // completed messages and track the current one separately.
   const stats = new QueryStats();
   const getStatusText = () => display.c.darkGray(stats.getStatusText());
   display.startStatus(getStatusText);
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -118,7 +118,7 @@ export async function runQuery(
   // completed messages and track the current one separately.
   const stats = new QueryStats();
   const workingVerb = display.pickWorkingVerb();
-  const getStatusText = () => stats.getStatusText(workingVerb);
+  const getStatusText = () => display.c.darkGray(stats.getStatusText(workingVerb));
   display.startStatus(getStatusText);
 
   const canUseTool: CanUseTool = async (toolName, input) => {
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -19,6 +19,7 @@ import type { ModelInfo, FetchModelsFn } from "./model.js";
 import { handleEffortCommand } from "./effort.js";
 import type { EffortValue } from "./effort.js";
 import { CommandRegistry } from "./command-registry.js";
+import { QueryStats } from "./query-stats.js";
 export { parseSlashCommand, dispatchInput, matchCommands, ask } from "./input.js";
 export type { SlashCommandResult, DispatchResult } from "./input.js";
 export { handleModelCommand, getCachedModels, _resetCachedModels } from "./model.js";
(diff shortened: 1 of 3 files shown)


src/agent/query-stats.ts
```diff
@@ -0,0 +1,66 @@
...
+    this.emit("change");
+  }
+
+  /** Formatted status bar text for use with display.startStatus(). */
+  getStatusText(workingVerb: string): string {
+    const secs = this.elapsedSecs;
+    const outTokens = this.outputTokens;
+    return display.c.darkGray(
+      `${workingVerb}… ${display.fmtStats(secs, this._turns || undefined, outTokens || undefined, this._inputTokens || undefined)}`,
+    );
```
jasoncrawford: I tend to think text styling (darkGray) should be done outside a module like this. I would have this method just return the formatted text, and have the caller wrap it in darkGray

jasoncrawford (commented):
Done —  now returns unstyled text; the  call is in the  caller.

jasoncrawford (commented):
Done — `getStatusText` now returns unstyled text; the `darkGray` call is in the `index.ts` caller.

src/agent/index.ts
```diff
@@ -112,17 +113,12 @@ export async function runQuery(
     process.stdout.write("\r\n\x1b[J");
   }
 
-  const startTime = Date.now();
   // Accumulate stats from stream_event messages to show in the status line.
   // message_delta.usage.output_tokens is cumulative per message, so we sum
   // completed messages and track the current one separately.
-  const stats = { turns: 0, inputTokens: 0, completedOutputTokens: 0, currentOutputTokens: 0 };
+  const stats = new QueryStats();
   const workingVerb = display.pickWorkingVerb();
```
jasoncrawford: let's move the choice of working verb, and the pickWorkingVerb method itself, into the new QueryStats? that would consolidate some responsibilities

jasoncrawford (commented):
Done — `QueryStats` now picks its own verb in the constructor, and `getStatusText()` takes no arguments.

src/agent/index.ts
```diff
@@ -112,17 +113,11 @@ export async function runQuery(
     process.stdout.write("\r\n\x1b[J");
   }
 
-  const startTime = Date.now();
   // Accumulate stats from stream_event messages to show in the status line.
   // message_delta.usage.output_tokens is cumulative per message, so we sum
   // completed messages and track the current one separately.
```
jasoncrawford: This comment contains details that are now only relevant to the internals of QueryStats

PR #637, opened 2026-04-12
fix: use id: undefined instead of id: 0 for unsaved WebhookEvent

### LINKED ISSUE #629: fix: use id: undefined instead of id: 0 for unsaved WebhookEvent
Small cleanup as part of the type system refactor (`docs/type-system.md`).

`WebhookEvent.fromIncoming()` currently passes `id: 0` as a placeholder when constructing an in-memory instance before DB insert. The design calls for `id: number | undefined` — an undefined id simply means the record is unsaved, which is a valid state.

This requires a small refactor to the private constructor to accept an optional id, or a separate in-memory-only constructor path that doesn't require a full `DbRow`.

jasoncrawford:
## Summary

- `WebhookEvent.id` is now `number | undefined` — `undefined` means the record is unsaved (not yet persisted to DB), which is the correct semantic vs. the previous sentinel `0`
- Constructor updated to accept `Omit<DbRow, "id"> & { id?: number }` so `fromIncoming()` can simply omit `id`
- `LogEntry.id` updated to `number | undefined` in both server (`admin-ws.ts`) and frontend (`frontend/src/types.ts`) types
- Added test asserting `WebhookEvent.fromIncoming()` produces `id === undefined`

## Test plan

- [ ] `npm test` passes (all 1313 non-DB tests pass)
- [ ] `npx tsc --noEmit` passes with no errors
- [ ] `npm run lint` passes

Closes #629

### DIFF
--- a/src/foreman/models/webhook-event.ts
+++ b/src/foreman/models/webhook-event.ts
@@ -8,7 +8,7 @@ type DbRow = Database["public"]["Tables"]["webhook_events"]["Row"];
 // ── Model ──────────────────────────────────────────────────────────────────────
 
 export class WebhookEvent {
-  readonly id: number;
+  readonly id: number | undefined;
   readonly deliveryId: string | null;
   readonly eventName: string;
   readonly action: string | null;
@@ -22,7 +22,7 @@ export class WebhookEvent {
   readonly payload: Record<string, unknown>;
   readonly receivedAt: string;
 
-  private constructor(row: DbRow) {
+  private constructor(row: Omit<DbRow, "id"> & { id?: number }) {
     this.id = row.id;
     this.deliveryId = row.delivery_id;
     this.eventName = row.event_name;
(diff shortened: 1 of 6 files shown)


src/foreman/models/webhook-event.ts
```diff
@@ -22,7 +22,7 @@ export class WebhookEvent {
   readonly payload: Record<string, unknown>;
   readonly receivedAt: string;
 
-  private constructor(row: DbRow) {
+  private constructor(row: Omit<DbRow, "id"> & { id?: number }) {
```
jasoncrawford: Feels it would be helpful to have a helper method/type constructor type thing somewhere, that given a `table`, does:

```Omit<Database["public"]["Tables"][table]["Row"], "id"> & { id?: number }```

Every db-backed active record model could use that helper.

jasoncrawford: Done — added `DbRow<T>` to `db-client.ts` (already imported by all models) and updated `WebhookEvent` to use `DbRow<"webhook_events">`.

PR #639, opened 2026-04-12
refactor: extract handleBusyHello and handleIdleHello from wss.ts

### LINKED ISSUE #620: Extract busy-hello reconnection logic from handleWorkerHello
## Background

`handleWorkerHello` in `src/foreman/controllers/wss.ts` is ~76 lines. The bulk of this is the "busy" reconnection branch — a 4-case decision tree over existing task state.

## Problem

The busy reconnection path is the most complex and failure-prone part of the worker handshake, but it currently lives inside a closure inside a WebSocket message handler, making it effectively untestable without spinning up a full WSS server and WebSocket connection.

## Proposed change

Extract the busy reconnection logic into a standalone function, e.g.:

```ts
async function handleBusyHello(
  workerId: string,
  claimedTaskId: string,
  deps: { registry, taskManager, sendMsg, log, flog }
): Promise<void>
```

This function handles the 4 cases:
1. Unknown task → create placeholder, reclaim
2. Complete task, same worker → reclaim for finalization
3. Complete task, different worker → cancel
4. Live task, different worker → cancel
5. Otherwise → reclaim

The idle-hello path (simple: revert prior task if any, register, ack) stays in place or becomes a small `handleIdleHello` helper.

## Benefit

Each reconnection case can be unit-tested by calling `handleBusyHello` directly with a mock task state, without standing up a WebSocket server.

jasoncrawford:
## Summary

- Extracts the busy reconnection logic from `handleWorkerHello` in `wss.ts` into two standalone exported functions in a new `src/foreman/controllers/hello-handlers.ts` module: `handleBusyHello` and `handleIdleHello`
- `handleWorkerHello` in `wss.ts` is now a thin 5-line wrapper that delegates to these helpers
- Adds 11 unit tests in `tests/foreman.hello-handlers.test.ts` covering all 5 reconnection cases without requiring a live WebSocket server

## Test plan

- [ ] All 11 new unit tests pass (`handleBusyHello` unknown task numeric/non-numeric, complete task same/different worker, live task same/different worker, queued event flush, `handleIdleHello` with/without prior task, revert failure logging)
- [ ] All 53 existing `foreman.websocket.test.ts` tests still pass

Closes #620


src/foreman/controllers/hello-handlers.ts
jasoncrawford, line 1: Does not feel like this should be a separate file. There is high coupling between this and wss.ts, and now each of them has lower cohesion. And there are a lot of deps passed between them.

wss.ts was not so large that it needed to be broken up. The goal was to break up a long method, not a long file.

jasoncrawford:
Good call. Moved `handleBusyHello` and `handleIdleHello` back into `wss.ts` as module-level exported functions and deleted `hello-handlers.ts`. The testability goal is preserved — tests still import and call them directly — without the cross-file coupling.

PR #659, opened 2026-04-13
refactor: centralize flog into src/utils.ts

### LINKED ISSUE #658: flog is redefined in several files
Feels like we should just have logging functions defined once in a logging file, or a utils file. Any module can import it just like a util, doesn't need to be passed around as a dependency or anything.

jasoncrawford:
## Summary

- `flog` was defined identically in three separate files (`foreman/index.ts`, `controllers/http-server.ts`, `controllers/wss.ts`) and also passed around as an injected dependency in `EventRouter`, `BusyHelloDeps`, `IdleHelloDeps`, and two `TaskManager` methods
- Moved the single definition to `src/utils.ts` and updated all callers to import it directly
- Removed `flog` from all constructor/function deps interfaces — no more passing it around
- Updated tests to use `vi.spyOn(utils, 'flog')` where logging assertions are needed

## Test plan

- [ ] All 1381 unit tests pass (`npm test`)

Closes #658

### DIFF
--- a/src/foreman/controllers/http-server.ts
+++ b/src/foreman/controllers/http-server.ts
@@ -6,11 +6,7 @@ import type { TaskManager } from "../models/task-manager.js";
 import { Task } from "../models/task.js";
 import { queryActivityLog } from "../models/activity-log.js";
 import type { TaskStatus } from "../../../shared/types.js";
-import { fmtError } from "../../utils.js";
-
-function flog(msg: string) {
-  console.log(`${new Date().toISOString()} ${msg}`);
-}
+import { fmtError, log } from "../../utils.js";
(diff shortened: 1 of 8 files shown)


jasoncrawford (commented):
While we're doing this, can we call it something other than "flog"? That's a dictionary word that means something else, and it's obscure. What is it supposed to mean, "foreman log"? It could just be called "log" frankly?

jasoncrawford (commented):
Good call — renamed to `log` in the latest commit. The local worker-prefix helper in `wss.ts` (which wraps `log` to prepend the worker ID) became `workerLog` to avoid shadowing.

PR #660, opened 2026-04-13
refactor: extract ActiveRecord base class for common CRUD boilerplate

### LINKED ISSUE #656: Extract ActiveRecord base class for common CRUD boilerplate
The foreman model layer (`src/foreman/models/`) has three active-record model classes: `Task`, `ForemanMessage`, and `WebhookEvent`. Each independently implements a set of CRUD boilerplate that belongs in a shared base class. This issue covers what can be extracted without code generation — the row↔model field mapping layer is out of scope and tracked separately.

## Background: what boilerplate exists today

Each model independently implements:

- A `db.from("table_name").select("*")` expression inlined in every query method (`Task` has a private `select()` helper; the others don't)
- Static single-record finders (`get(id)`, lookups by other columns)
- A `query()` / `list()` method
- `delete()`
- `insert()` or `upsert()`
- In `Task`: a `save()` method that calls `db.update()` and emits `Task.events.emit("changed")`

Additionally, `ForemanMessage.log()` and `WebhookEvent.log()` are currently declared `void` and fire-and-forget. **These should return `Promise<void>`** so callers can `await` if needed and use `void log(...)` at call sites where they don't care about completion. This is a design improvement independent of the base class work but should be done in the same PR.

## What to build

An abstract `ActiveRecord` base class in `src/foreman/models/active-record.ts`.

### Each subclass declares

```ts
protected static readonly tableName: string;   // e.g. "foreman_messages"
protected static readonly primaryKey: string;  // e.g. "id" or "task_id"
```

### Base class provides

**`static readonly events: EventEmitter`**
Each subclass gets its own event emitter (declared on the subclass, not shared). The base class `update()` method calls `(this.constructor as typeof ActiveRecord).events.emit("changed")` after every successful update. If no listeners are subscribed, the emit is a no-op — no special casing needed. `Task` currently uses `Task.events` for this; that moves to the base pattern and works identically.

**`protected static select()`**
Returns `db.from(this.tableName).select("*")`. All query methods call this instead of inlining `db.from(...)`.

**`static async get(id): Promise<InstanceType<this> | null>`**
Finds by primary key: `select().eq(primaryKey, id).maybeSingle()`. Throws on error. Wraps result in `new this(row)`. Replaces the private `findOne()` in `Task` and makes it available everywhere.

**`static async getBy(col, val): Promise<InstanceType<this> | null>`**
Same, but with an arbitrary column.

**`static async list(opts?: { limit?: number }): Promise<InstanceType<this>[]>`**
Selects all with optional limit. Each model's current `query(opts)` / `list(opts)` becomes this. Subclasses can override to add default ordering.

**`static async insert(data): Promise<InstanceType<this>>`**
Wraps `db.from(tableName).insert(data).select().single()` and returns the inserted record as a model instance. This ensures server-generated fields (e.g. `id`, `created_at`) are reflected in the returned object. Throws on error.

**`protected async update(changes): Promise<InstanceType<this>>`**
Calls `db.from(tableName).update(changes).eq(primaryKey, this[primaryKey]).select().single()`, emits `"changed"` on the subclass's event emitter, and returns a new model instance reflecting the updated row. Callers (e.g. `Task.assign()`) can use the returned instance to refresh local state. Throws on error.

**`async delete(): Promise<void>`**
Calls `db.from(tableName).delete().eq(primaryKey, this[primaryKey])`.

## TypeScript notes

- `db.from(tableName)` where `tableName` is a runtime string loses Supabase's per-table type inference. Use `as any` / `as unknown` casts internally — this is acceptable inside the base class. Type the return values explicitly on public methods.
- Polymorphic static methods that return `InstanceType<this>` require a constructor type pattern. Study how `Task.findOne()` currently works and extend the approach. The goal is correct types in subclasses, not perfectly abstract generics — cast where needed.
- Don't over-engineer. The primary goal is eliminating runtime boilerplate; a few `as any` casts in the base internals are fine.

## What does NOT change in this issue

- Property declarations and constructors remain per-model (addressed in the codegen issue)
- `toJSON()` remains per-model
- `fromTest()` remains per-model
- Semantic methods (`Task.assign()`, `Task.complete()`, `ForemanMessage.buildSummary()`, etc.) remain per-model
- `Task.upsert()` has non-trivial `ON CONFLICT` logic — keep it in `Task`, calling the base `insert()` or a similar primitive
- Field sync in `Task.save()` (the `if ("worker_id" in changes) this.workerId = ...` block) remains per-model for now; `update()` returns the refreshed instance which the subclass can use to sync properties, but the per-field sync logic itself is not changed until the codegen issue is done

## Testing

`npm test` must pass. The existing test suite exercises the model layer via `setupInMemoryTasks()` and mocks at the `Task` method level — no test changes should be required. Run `npx tsc --noEmit` as well.

jasoncrawford:
## Summary

- Adds abstract `ActiveRecord` base class in `src/foreman/models/active-record.ts` with shared CRUD methods: `select()`, `get()`, `getBy()`, `list()`, `insert()`, `update()`, `delete()`
- `Task`, `ForemanMessage`, and `WebhookEvent` all extend `ActiveRecord`, eliminating duplicated `db.from(tableName).select("*")` boilerplate and independent finder patterns
- `Task.save()` delegates to base `update()` (which handles the DB call and emits `"changed"`); per-field property sync remains in `Task` for now
- `ForemanMessage.query()` and `WebhookEvent.query()` renamed to `list()` to match the base class interface; `activity-log.ts` caller updated accordingly
- Docs updated: `CLAUDE.md` and `docs/type-system.md` now mention `ActiveRecord`

## TypeScript notes

The base class uses `as any` casts internally (per the issue spec) because `db.from(tableName)` with a runtime string loses Supabase's per-table type inference. Subclass methods that override base methods provide concrete return types. Type check (`npx tsc --noEmit`) passes cleanly.

## Test plan

- [x] `npm test` — all 1381 unit tests pass (3 DB tests skipped — require `supabase start`, pre-existing)
- [x] `npx tsc --noEmit` — no errors

Closes #656


src/foreman/models/active-record.ts
jasoncrawford, line 1: do we need so many `any`s, and the eslint disables? e.g., can't get(id) return Promise<this>?

src/foreman/models/task.ts
jasoncrawford: is this still needed given that we have a delete on the superclass now? is it because of the assigned_at check? maybe we should rename this method if it's not an unconditional delete?

src/foreman/models/webhook-event.ts
jasoncrawford: can this be done in a different way? like, create a model object an then call save on it?

src/foreman/models/foreman-message.ts
jasoncrawford: can this be done in a different way? like, create a model object an then call save on it?

src/foreman/models/active-record.ts
jasoncrawford: Reduced to a single file-level `/* eslint-disable @typescript-eslint/no-explicit-any */` at the top, which is cleaner than per-line suppresses. The `any`s inside the method bodies are intentional — since `tableName` is a runtime string, Supabase can't infer per-table types, so we cast internally. As for `get(id): Promise<this>` — `this` in a static method refers to the class, not the instance, so it's not directly usable for the return type. I tried `this: T extends typeof ActiveRecord` with `InstanceType<T>`, but TypeScript rejects it because `private constructor(row: DbRow)` makes `typeof Task` incompatible with `typeof ActiveRecord` (constructor signature mismatch). The `any`-returning base with concrete-typed overrides is the cleanest approach that actually compiles.

src/foreman/models/task.ts
jasoncrawford: Yes — it's the `assigned_at` guard that differentiates it from the base class `delete()`. Renamed to `deleteIfUnassigned()` to make the conditional semantics explicit. Callers in `task-manager.ts` and `github.ts` updated accordingly, and the tests updated to call the new name.

src/foreman/models/webhook-event.ts
jasoncrawford: Good call. The `log()` methods now delegate to `ActiveRecord.insert()` instead of calling `db.from()` directly. The fire-and-forget semantics are preserved — `insert()` throws on error, and `log()` catches it with `.catch()`. The `db` import is no longer needed in these files and was removed.

src/foreman/models/foreman-message.ts
jasoncrawford: Good call. The `log()` methods now delegate to `ActiveRecord.insert()` instead of calling `db.from()` directly. The fire-and-forget semantics are preserved — `insert()` throws on error, and `log()` catches it with `.catch()`. The `db` import is no longer needed in these files and was removed.

src/foreman/models/active-record.ts
jasoncrawford: OK. Add a comment to the eslint directive about why we are allowing explicit-any in this file?

jasoncrawford: Added a 3-line comment right below the directive explaining both reasons.

PR #666, opened 2026-04-14
refactor: pass model objects instead of IDs in foreman internals

### LINKED ISSUE #655: Pass model objects, not ids
General pattern we should follow:

* controllers parse ids as params out of messages
* controllers look up model objects by id using finder methods, first thing
* after that, methods pass model objects around, not ids

E.g., Worker has an assign() method that takes a taskId; instead it should take a task and pull the id off it as needed. Or the EventQueue, which takes taskIds. Or the methods on TaskManager that call the eventQueue. Etc.

## Implementation notes

The following is a rough guide based on the code at time of writing. By the time you pick this up, things may have changed — treat this as a starting point, and do your own audit for other places where the same pattern applies.

**`Worker` (`src/foreman/models/worker-registry.ts`)**
- `currentTaskId?: string` → `currentTask?: Task`
- `assign(taskId: string)` → `assign(task: Task)`
- `release()` — clears `currentTask` instead of `currentTaskId`
- `toSnapshot()` — derives `currentTaskId: this.currentTask?.taskId` for the wire type
- `getByTask(taskId: string)` — keep as-is; static finders taking IDs are correct by definition

**`EventQueue` (`src/foreman/event-queue.ts`)**
- `enqueue(taskId, event)` → `enqueue(task: Task, event)` — use `task.taskId` as the internal map key
- `drain(taskId)` → `drain(task: Task)`

**`TaskManager` (`src/foreman/models/task-manager.ts`)**
- `queueEvent(taskId, event)` → `queueEvent(task: Task, event)`
- `drainEvents(taskId)` → `drainEvents(task: Task)`
- `registerBranch(branch, taskId)` → `registerBranch(branch, task: Task)` — keep storing `task.taskId` as the map key internally; `getTaskForBranch()` should continue doing a fresh `Task.get()` lookup rather than returning a potentially stale stored reference
- `loadActiveTasksFromDb` directly sets `branchToTaskId` — update to call `registerBranch()` instead

**`wss.ts` (`src/foreman/controllers/wss.ts`)**
- `flushQueuedEvents(taskId, issueRef)` → `flushQueuedEvents(task: Task)` — derive both from the Task
- `worker.assign(task.taskId)` → `worker.assign(task)` (appears in multiple places)

**`event-router.ts` (`src/foreman/controllers/event-router.ts`)**
- `taskManager.queueEvent(task.taskId, evt)` → `taskManager.queueEvent(task, evt)`
- `taskManager.registerBranch(branch, linkedTask.taskId)` → `taskManager.registerBranch(branch, linkedTask)`
- `worker.currentTaskId !== task.taskId` → `worker.currentTask?.taskId !== task.taskId`

**What doesn't change:** `enqueueIssue(taskId, issueNumber, ...)` in TaskManager — the task doesn't exist yet at that point (being created by the upsert inside), so an ID is correct there. Similarly, anywhere that extracts `task.taskId` to send over the wire is serialization, not internal ID-passing.

Run `npm test` and `npx tsc --noEmit` to verify.

jasoncrawford:
## Summary

- `Worker.assign()` now takes a `Task` object; `currentTask` replaces `currentTaskId` as the stored field. A `currentTaskId` getter is kept for backward-compatible wire serialization (`toWire()`).
- `EventQueue.enqueue()` and `drain()` take a `Task` instead of a raw `taskId` string.
- `TaskManager.queueEvent()`, `drainEvents()`, and `registerBranch()` take a `Task`. `loadActiveTasksFromDb` now calls `registerBranch()` instead of writing to `branchToTaskId` directly.
- `wss.ts`: `flushQueuedEvents()` takes a `Task` (derives both `taskId` and `issueNumber` from it); `reclaimWorker()` and `forwardEvent()` pass Task objects through.
- All call sites in tests updated to supply Task objects.

## Test plan

- [x] `npm test` — all 1388 unit tests pass (3 DB tests skipped as expected without Supabase)
- [x] `npx tsc --noEmit` — no type errors

Closes #655


src/foreman/controllers/wss.ts
```diff
@@ -284,16 +288,16 @@ export class ForemanWss {
...
+    this.flog(`[worker ${shortWorkerId(wid)}] ${line}`);
   }
 
   // ── Hello handlers ───────────────────────────────────────────────────────────
 
-  private flushQueuedEvents(workerId: string, taskId: string, issueRef: string | number): void {
-    for (const evt of this.taskManager.drainEvents(taskId)) {
-      this.sendMsg(workerId, { type: "event_notification", taskId, event: evt.toWorkerPayload() });
-      this.workerLog(workerId, `→ event_notification #${issueRef} ${evt.eventName} (queued)`);
+  private flushQueuedEvents(workerId: string, task: Task): void {
```
jasoncrawford: why doesn't this take a worker model object, too? or just get the workerId off the task object?

src/foreman/controllers/wss.ts
```diff
@@ -304,14 +308,14 @@ export class ForemanWss {
 
   private async reclaimWorker(workerId: string, task: Task, ws: WebSocket): Promise<void> {
     const w = Worker.register(workerId, ws);
-    w.assign(task.taskId);
+    w.assign(task);
     // Only call assign if task is not already complete (to preserve task status)
     if (task.status !== "complete") {
       await task.assign(workerId);
     }
```
jasoncrawford, line 310: why does the worker have to assign the task, and the task also assign the worker?

once again we're passing a raw id into a method

src/foreman/models/task-manager.ts
```diff
@@ -82,10 +81,10 @@ export class TaskManager extends EventEmitter {
   private async tryAssignWork(workerId: string): Promise<AssignOutcome | null> {
```
jasoncrawford: why doesn't this take a worker model object?

jasoncrawford (commented):
Good changes, see a few specific comments on more we could do

src/foreman/models/task-manager.ts
jasoncrawford: Done — `tryAssignWork` now takes a `Worker` object directly. `assignIdleWorkers` passes `w` (the Worker object from `Worker.getIdle()`) straight through.

src/foreman/controllers/wss.ts
jasoncrawford: Good question. `worker.assign(task)` updates in-memory Worker state (sets `currentTask`, flips status to `busy`, emits `changed`). `task.assign(worker.workerId)` persists the assignment to the database (writes `worker_id` and `assigned_at` to the DB row). They do different things at different layers — the Worker model doesn't touch the DB, and the Task's DB method needs the ID string as the value to store. So both calls are necessary, and `worker.workerId` is the right way to get that string from the object.

src/foreman/controllers/wss.ts
jasoncrawford: Done — `flushQueuedEvents` now takes a `Worker` object. `reclaimWorker` still takes `workerId: string` at its entry point (that's what arrives from the wire protocol and is needed for `Worker.register(workerId, ws)`), but all subsequent calls inside use `w` (the Worker object returned by register) rather than the raw `workerId` parameter.

src/foreman/models/task-manager.ts
```diff
@@ -70,25 +70,25 @@ export class TaskManager extends EventEmitter {
...
-  private async tryAssignWork(workerId: string): Promise<AssignOutcome | null> {
+  private async tryAssignWork(worker: Worker): Promise<AssignOutcome | null> {
     const task = await this.nextPending(t => t.blockersLoaded && t.status === "pending");
     if (!task) return null;
-    Worker.get(workerId)?.assign(task.taskId);
+    worker.assign(task);
     try {
-      await task.assign(workerId);
-      return { ok: true, task, queued: this.drainEvents(task.taskId), workerId };
+      await task.assign(worker.workerId);
```
jasoncrawford: why does task.assign still take a workerId? can it take a worker model object?

src/foreman/models/task-manager.ts
```diff
@@ -70,25 +70,25 @@ export class TaskManager extends EventEmitter {
...
+      return { ok: true, task, queued: this.drainEvents(task), workerId: worker.workerId };
```
jasoncrawford: this object is purely internal, right? not serialized over the wire... so why can't this just have the worker model, instead of a workerId?

src/foreman/models/task-manager.ts
jasoncrawford: Done. `task.assign()` now takes `{ workerId: string }` — Worker satisfies this structurally so no circular import is needed, and tests can pass `{ workerId: "w1" }` inline without needing a real Worker instance. All call sites updated.

src/foreman/models/task-manager.ts
jasoncrawford: Done. `AssignOutcome` now carries `worker: Worker` instead of `workerId: string`. Both success and failure branches updated, and `assignWork()` in `wss.ts` now uses `outcome.worker.workerId` throughout.

src/foreman/controllers/wss.ts
```diff
@@ -304,14 +304,14 @@ export class ForemanWss {
 
   private async reclaimWorker(workerId: string, task: Task, ws: WebSocket): Promise<void> {
```
jasoncrawford: can this method take a worker instead of a workerId?

src/foreman/controllers/wss.ts
```diff
@@ -304,14 +304,14 @@ export class ForemanWss {
...
+    w.assign(task);
     // Only call assign if task is not already complete (to preserve task status)
     if (task.status !== "complete") {
-      await task.assign(workerId);
+      await task.assign(w);
     }
     // For complete tasks, the task stays complete while worker finishes cleanup/finalization work
-    this.sendMsg(workerId, { type: "hello_ack", workerId, status: "busy" }, task.taskId);
-    this.flushQueuedEvents(workerId, task.taskId, task.issueNumber);
+    this.sendMsg(w.workerId, { type: "hello_ack", workerId: w.workerId, status: "busy" }, task.taskId);
```
jasoncrawford: can sendMsg take a worker instead of an id?

src/foreman/models/task.ts
```diff
@@ -208,8 +208,8 @@ export class Task extends ActiveRecord {
 
   // ── Instance mutations ──────────────────────────────────────────────────────
 
-  async assign(workerId: string): Promise<void> {
-    await this.save({ worker_id: workerId, assigned_at: new Date().toISOString() });
+  async assign(worker: { workerId: string }): Promise<void> {
```
jasoncrawford: why is this type `{ workerId: string }` and not type `Worker`?

src/foreman/controllers/wss.ts
jasoncrawford: Done. `Worker.register` is now called once at the top of `handleBusyHello` (the controller entry point, where the wire message is parsed), and the resulting Worker object is passed to both `reclaimWorker` and `cancelWorker`. Both helpers now take `Worker` instead of `workerId + ws`.

src/foreman/controllers/wss.ts
jasoncrawford: Done. `sendMsg(worker: Worker, ...)` now takes a Worker directly — calls `worker.send(msg)` instead of `Worker.get(workerId)?.send(msg)`, eliminating the redundant lookup. All call sites updated.

src/foreman/models/task.ts
jasoncrawford: The structural type `{ workerId: string }` was chosen so tests can pass `{ workerId: "w1" }` without needing to register a real Worker (which would add registry state to every test that sets up a task). Worker satisfies the type structurally, so production code is unaffected. Happy to switch to `Worker` with `import type` if you'd prefer the explicit type — the tradeoff is that tests would need `Worker.register` calls everywhere task state is seeded.

src/foreman/controllers/wss.ts
```diff
@@ -290,28 +291,26 @@ export class ForemanWss {
...
+    for (const evt of this.taskManager.drainEvents(task)) {
+      this.sendMsg(worker, { type: "event_notification", taskId: task.taskId, event: evt.toWorkerPayload() });
+      this.workerLog(worker.workerId, `→ event_notification #${task.issueNumber} ${evt.eventName} (queued)`);
     }
   }
 
-  private cancelWorker(workerId: string, taskId: string | undefined, ws: WebSocket): void {
-    Worker.register(workerId, ws);
-    this.sendMsg(workerId, { type: "hello_ack", workerId, status: "cancelled" }, taskId);
+  private cancelWorker(worker: Worker, taskId: string | undefined): void {
```
jasoncrawford: is there a Task object for this? if so pass it instead

src/foreman/controllers/wss.ts
```diff
@@ -327,6 +326,7 @@ export class ForemanWss {
    */
   async handleBusyHello(workerId: string, claimedTaskId: string, ws: WebSocket): Promise<void> {
     const existing = await Task.get(claimedTaskId);
+    const w = Worker.register(workerId, ws);
```
jasoncrawford: minor stylistic thing, but I prefer to spell out variables like `worker`

src/foreman/models/task.ts
jasoncrawford: Yeah seems better to use real model objects; fine to have more setup in tests if it's just creating objects

jasoncrawford: Done — `task.assign(worker: Worker)` now uses `import type { Worker }` (erased at runtime, no circular dep). Tests updated to use `Worker.register(id, fakeWs)` where task state was previously seeded with `{ workerId }` structs.

src/foreman/controllers/wss.ts
jasoncrawford: Done. `cancelWorker` now takes `Task | null`. Callers pass the `existing` Task object where one was already looked up, or `null` for the non-numeric unknown-task case (where there genuinely is no Task). The `taskId` log entry uses `task?.taskId`.

src/foreman/controllers/wss.ts
jasoncrawford: Done — renamed `w` to `worker` throughout `handleBusyHello`.

PR #673, opened 2026-04-14
refactor: move agent ID generation into worker.ts

### LINKED ISSUE #672: Move agent ID generation into worker.ts
WORKER_NAMES and generateAgentId are in src/utils.ts but they're only relevant to the worker.

Move them into src/agent/worker.ts. Instead of having main() in src/agent/index.ts generate the agent ID, move that into worker.ts also. The AgentStatus can just do it in its constructor. Then WORKER_NAMES can be private, doesn't need to be exported.

jasoncrawford:
## Summary

- Moved `WORKER_NAMES` and `generateAgentId` from `src/utils.ts` into `src/agent/worker.ts`, where they are only relevant
- `WORKER_NAMES` is now private (not exported)
- `AgentStatus` generates its own `agentId` in the constructor via a default parameter (`agentId = generateAgentId()`), so callers no longer need to generate and pass one
- Removed the `generateAgentId()` call from `main()` in `src/agent/index.ts`
- Updated `tests/utils.test.ts` to import `generateAgentId` from `src/agent/worker.js` and removed the `WORKER_NAMES` reference (replaced with a regex check)

## Test plan

- [x] All 1389 unit tests pass
- [x] `npx tsc --noEmit` clean

Closes #672


src/agent/index.ts
```diff
@@ -232,9 +232,7 @@ export async function main(
   initialModel?: string,
   initialEffort?: EffortValue,
 ): Promise<void> {
-  // Generate agentId unconditionally — used as both workspace identity and foreman identity.
-  const agentId = generateAgentId();
-  const agentStatus = new AgentStatus(agentId, { model: initialModel, effort: initialEffort });
+  const agentStatus = new AgentStatus(undefined, { model: initialModel, effort: initialEffort });
```
jasoncrawford: kind of ugly to pass undefined here, we could just omit this argument since this is the only place it's ever called? or if we really want it, make it an optional attribute in the options hash

jasoncrawford (commented):
Good catch. Restructured the constructor to take a single options hash `{ agentId?, ...initial }` so the call site is now `new AgentStatus({ model: initialModel, effort: initialEffort })` — no `undefined` needed. Updated all test call sites too (e.g. `new AgentStatus({ agentId: AGENT_ID })`).

PR #681, opened 2026-04-15
refactor: extract status bar into status-bar.ts reactive module

### LINKED ISSUE #674: Extract status bar into its own reactive module
The status bar is currently rendered as part of `display.ts` (a 1000+ line file). It should be its own module that owns its state and renders itself reactively.

**Fix:**
- Extract status bar logic from `display.ts` into `status-bar.ts` (or similar)
- The module subscribes to relevant model events (worker status, current task, token counts, etc.) and re-renders on change
- This follows the reactive-UI pattern described in CLAUDE.md: a model holds state and emits on change; the UI subscribes once rather than scattering manual refresh calls
- `display.ts` is left with its remaining concerns: color/style constants, print helpers, diff rendering — which may still be worth splitting further, but is much more manageable without the status bar

jasoncrawford:
## Summary

- Extracts the verbose flag, `fmtWorkerStatus` formatter, status bar state/render functions, and input callbacks from `display.ts` into a new `src/agent/status-bar.ts` module
- `display.ts` re-exports all moved symbols for full backward compatibility — no callers need to change their imports
- `display.ts`'s `print()` now calls `clearStatusBars()`/`drawStatusBars()` from `status-bar.ts`; `printBlock()` calls `fireOnToolResult()` instead of directly calling the callback variable
- Added `tests/status-bar.test.ts` with 22 tests covering the new module's direct API
- Updated `CLAUDE.md` to document the new file

**Line counts:** `display.ts` 1012 → 809 lines; new `status-bar.ts` is 261 lines

The `AgentStatus` reactive model (in `worker.ts`) already emits `"change"` events, and `WorkerSession.start()` subscribes once — this PR completes the picture by giving the status bar its own focused module as described in the reactive-UI pattern in CLAUDE.md.

## Test plan

- [x] All 1430 existing tests pass
- [x] 22 new `status-bar.test.ts` tests pass
- [x] `npx tsc --noEmit` reports no type errors

Closes #674


jasoncrawford (commented):
Two patterns here need to be corrected before this merges.

## 1. `status-bar.ts` should be a class, not a bag of module-level functions

`status-bar.ts` moves state out of `display.ts` but keeps it as module-level `let` variables with exported functions that close over them. This is an implicit singleton — it has all the downsides of global state (hidden, untestable, only one instance possible) with none of the clarity of a class.

All of this:
```typescript
let _statusText = "";
export let _statusActive = false;
let _statusInterval: ReturnType<typeof setInterval> | null = null;
let _inputPrintCallback: (() => void) | null = null;
export function setInputPrintCallback(fn: ...) { _inputPrintCallback = fn; }
export function startStatus(getText: () => string) { ... }
export function stopStatus() { ... }
```

Should be a class:
```typescript
export class StatusBar {
  private statusText = "";
  active = false;
  private interval: ReturnType<typeof setInterval> | null = null;
  private inputPrintCallback: (() => void) | null = null;

  setInputPrintCallback(fn: ...) { this.inputPrintCallback = fn; }
  start(getText: () => string) { ... }
  stop() { ... }
}

export const statusBar = new StatusBar(); // one shared instance, clearly declared
```

Callers use `statusBar.start(...)`, `statusBar.stop()`, etc. The singleton is explicit and named, not hidden in module scope. The state is encapsulated, not scattered across `let` declarations.

## 2. The re-exports in `display.ts` should be removed; callers should be updated instead

```typescript
// display.ts — this entire block should not exist
export {
  verbose, setVerbose,
  WorkerStatusOpts, fmtWorkerStatus,
  startStatus, stopStatus,
  ...
} from "./status-bar.js";
```

Re-exports for backward compatibility defeat the purpose of the split. `display.ts` still appears to "contain" everything it did before — the old coupling is intact, just hidden one layer deeper.

The right move: find every file that imports the moved symbols from `display.ts` and update those imports to point at `status-bar.ts` directly. `grep -r "from.*display" src/` will find them. Update all of them. There should be zero re-exports when you are done.

See the `no-re-export` and `module-design` skills for the full rationale.

src/agent/index.ts
```diff
@@ -6,7 +6,8 @@ import { fileURLToPath } from "url";
 import { query } from "@anthropic-ai/claude-agent-sdk";
 import type { CanUseTool, PermissionMode, PermissionResult } from "@anthropic-ai/claude-agent-sdk";
 import * as display from "./display.js";
-import { setVerbose, setThinkOutLoud } from "./display.js";
+import { setThinkOutLoud } from "./display.js";
```
jasoncrawford: thinkOutLoud could also come directly off of config

src/agent/status-bar.ts
```diff
@@ -0,0 +1,252 @@
...
+  branch?: string;
+  connectionStatus: "connected" | "disconnected" | "reconnecting" | "handshaking";
+  /** WebSocket close code, shown in verbose mode when disconnected. */
+  disconnectCode?: number;
+  /** Seconds until reconnect attempt, shown when disconnected. */
+  retryInSeconds?: number;
+  width?: number;
+}
+
+export function fmtWorkerStatus(opts: WorkerStatusOpts): string {
```
jasoncrawford: why isn't this just a method on StatusBar? if it were, I think it would eliminate this entire WorkerStatusOpts type, couldn't you get all these things off the object directly?

src/agent/worker.ts
```diff
@@ -735,10 +736,10 @@ export async function startWorkerMode(config: WorkerModeConfig, agentStatus: Age
...
     print: display.print,
     printForemanMessage: display.printForemanMessage,
-    startPersistentStatus: display.startPersistentStatus,
-    stopPersistentStatus: display.stopPersistentStatus,
-    updatePersistentStatus: display.updatePersistentStatus,
-    setOnToolResultCallback: display.setOnToolResultCallback,
+    startPersistent: (getText) => statusBar.startPersistent(getText),
+    stopPersistent: () => statusBar.stopPersistent(),
+    updatePersistent: () => statusBar.updatePersistent(),
+    setOnToolResult: (fn) => statusBar.setOnToolResult(fn),
```
jasoncrawford: why do we need these little methods that just forward calls to the status bar? why not replace all four with an instance of the status bar, and then call methods on statusBar directly?

jasoncrawford (commented):
I don't think verbose should live in status-bar.ts. If it shouldn't be in display.ts either, then maybe we should just get it out of config directly?

jasoncrawford:
Good feedback — here's what I'll do for each:

1. **`verbose`**: Making it a field on `StatusBar` (public getter + `setVerbose()` method) rather than a standalone module export. `display.ts` already imports `statusBar` so it reads `statusBar.verbose` directly — no circular import, no new module needed. 

2. **`fmtWorkerStatus` as method**: Absorbing `AgentStatus` entirely into `StatusBar`. `StatusBar` gains the worker-status fields (`agentId`, `connectionStatus`, `model`, etc.) and `update(patch)` from `AgentStatus`. `fmtWorkerStatus` becomes a private method reading from `this`, `WorkerStatusOpts` type is deleted, and `startPersistent()` becomes parameterless. `AgentStatus` class is removed from `worker.ts`.

3. **Wrapper methods**: `WorkerSession` constructor takes `StatusBar` directly as its first arg (renamed from `agentStatus`). Since it already has the instance, `WorkerDisplay` shrinks to just `{ print, printForemanMessage }` — the four status-bar methods vanish entirely.

src/agent/status-bar.ts
```diff
@@ -0,0 +1,353 @@
...
+  }
+
+  setInputPrint(fn: (() => void) | null): void { this._inputPrint = fn; }
+  getInputPrint(): (() => void) | null { return this._inputPrint; }
+
+  setInputStatus(fn: (() => void) | null): void { this._inputStatus = fn; }
+  getInputStatus(): (() => void) | null { return this._inputStatus; }
+
+  setInputClear(fn: (() => void) | null): void { this._inputClear = fn; }
+  getInputClear(): (() => void) | null { return this._inputClear; }
```
jasoncrawford: why not make these properties with real getters/setters? or for that matter, given the implementation here, why not just make them public properties?

jasoncrawford (commented):
Good call — making them public fields.

jasoncrawford (commented):
Thanks for the confirmation!

src/agent/status-bar.ts
```diff
@@ -0,0 +1,344 @@
...
+  "temperance", "thaddeus", "thankful", "theodore", "theophilus", "titus",
+  "tobias", "verity", "victor", "violet", "warren", "zephaniah",
+];
+
+/** Generate a human-readable agent ID by prepending a random human name to a UUID.
+ * E.g. "patience-a9bdda00-1234-5678-abcd-ef0123456789" */
+export function generateAgentId(): string {
+  const idx = randomInt(WORKER_NAMES.length);
+  return `${WORKER_NAMES[idx]}-${crypto.randomUUID()}`;
+}
```
jasoncrawford: ID generation doesn't seem like a status bar concern, seems more like a worker concern. I would keep this in worker.ts

src/agent/status-bar.ts
```diff
@@ -0,0 +1,344 @@
...
+  private _branch = "";
+  private _model: string | undefined;
+  private _effort: EffortValue | undefined;
+  private _countdownTimer: ReturnType<typeof setInterval> | null = null;
+
+  // ── Verbose flag ──────────────────────────────────────────────────────────
+  /** Whether verbose output mode is active (affects timestamps, detail level). */
+  verbose = false;
+
+  setVerbose(v: boolean): void { this.verbose = v; }
```
jasoncrawford: I don't think status bar should own verbose. Since this is a config option that doesn't change, just import the config global/singleton and get the verbose option off of there, rather than having it live here.

src/agent/status-bar.ts
```diff
@@ -0,0 +1,344 @@
...
+  updatePersistent(): void {
+    if (!this.persistentActive) return;
+    this.clear();
+    this._persistentText = this._fmtWorkerStatus();
+    this.draw();
+  }
+}
+
+/** Shared singleton — use this in production code. */
+export const statusBar = new StatusBar();
```
jasoncrawford: I don't love this, feels like the main loop in index.ts should instantiate this and pass it to whoever needs it

jasoncrawford:
Done — addressed all three in the latest commit:

**verbose**: Moved to a new `src/agent/verbose.ts` module (`export let verbose` + `setVerbose`). Both `display.ts` and `status-bar.ts` import from there, and `index.ts`/`worker.ts` call `setVerbose(config.verbose)` at startup. `workspace.ts` imports `verbose` directly too.

**generateAgentId / WORKER_NAMES**: Moved to `worker.ts`. The trick was that moving it there while `status-bar.ts` still defaulted `agentId` via `generateAgentId()` would create a circular dep (`worker.ts` → `status-bar.ts` → `worker.ts`). Resolved by making `agentId` required in the `StatusBar` constructor — no more self-generation.

**singleton**: `status-bar.ts` now exports `initStatusBar(sb)` and initialises the singleton with a plain UUID at module load. `index.ts` calls `initStatusBar(new StatusBar({ agentId: generateAgentId() }))` early at startup. ES module live bindings mean all importers (`display.ts`, `input.ts`, etc.) see the updated singleton automatically — no signature changes needed in those modules.

jasoncrawford (commented):
Thanks for the iteration. A few things still need addressing before this merges.

## Stop wrestling with `verbose`

`verbose` and `thinkOutLoud` do not belong in `status-bar.ts`, a two-line module, or anywhere new. Leave them exactly where they are in `display.ts` for now. They will be cleaned up as part of issue #691, which will add a `getConfig()` singleton to `config.ts` — at that point they become one-liners (`getConfig().verbose`) and the whole problem goes away. This PR does not need to solve that.

The circular import concern (status-bar needing `verbose` from display) is also resolved by #691: once `status-bar.ts` can call `getConfig().verbose` directly, it does not need to import anything from `display.ts`. Circular imports between the two modules would be fine in ES modules anyway (live bindings, values only used inside function bodies), but it becomes moot either way.

## `status-bar.ts` still needs to be a class

The module-level `let` variables and exported functions are the implicit singleton anti-pattern described in the `module-design` skill. The state (`_statusActive`, `_statusText`, `_statusInterval`, the callbacks, etc.) and the functions that operate on it should be a class:

```typescript
export class StatusBar {
  private statusText = "";
  active = false;
  private interval: ReturnType<typeof setInterval> | null = null;
  private inputPrintCallback: (() => void) | null = null;
  // ...

  start(getText: () => string) { this.active = true; ... }
  stop() { this.active = false; ... }
  setInputPrintCallback(fn: ...) { this.inputPrintCallback = fn; }
}

export const statusBar = new StatusBar();
```

Callers use `statusBar.start(...)`, `statusBar.stop()`, etc. The singleton is explicit and named rather than hidden in module scope.

## The re-exports in `display.ts` must go

```typescript
// This entire block should not exist
export {
  verbose, setVerbose,
  startStatus, stopStatus,
  ...
} from "./status-bar.js";
```

Find every file that imports the moved symbols from `display.ts` and update those imports to point at `status-bar.ts` directly. `grep -r "from.*display" src/` will find them. There should be zero re-exports when you are done. See the `no-re-export` skill.

jasoncrawford:
Done:

**verbose**: Removed `verbose.ts` — `verbose` and `setVerbose` now live in `display.ts` alongside `thinkOutLoud`. `status-bar.ts` imports `verbose` from `display.ts` (the circular dep is a non-issue as you noted). All callers updated.

**StatusBar class**: Agreed — the class is exactly the right pattern; the objection was to `verbose.ts` having module-level state + functions.

**Re-exports**: There are none — `display.ts` imports from `status-bar.ts` but doesn't re-export anything from it.

src/agent/workspace.ts
jasoncrawford, line 1: no need to import verbose separately; this file can be reverted to its original version

src/agent/status-bar.ts
```diff
@@ -0,0 +1,316 @@
...
+    this.draw();
+  }
+}
+
+/**
+ * Shared singleton. Initialized with a plain UUID at module load; call
+ * initStatusBar() early in startup to replace it with one that has a
+ * human-readable agent ID.
+ */
+export let statusBar: StatusBar = new StatusBar({ agentId: crypto.randomUUID() });
```
jasoncrawford: really no need to initialize it to a fake value; index.ts will initialize it immediately

jasoncrawford:
Done: **workspace.ts** reverted to use `display.verbose` via the existing `display` namespace (dropped the separate import). **statusBar singleton** now initialized with `undefined!` — tests get a valid instance via a new `initStatusBar()` call added to `tests/setup.ts`.

PR #692, opened 2026-04-15
refactor: merge effort.ts and model.ts into settings.ts

### LINKED ISSUE #689: Merge effort.ts and model.ts into settings.ts
Both `effort.ts` and `model.ts` in `src/agent/` are picker UIs for a single config value each — effort level and model selection respectively. They are the same kind of thing and don't warrant separate files.

**Fix:** Merge both into `src/agent/settings.ts` and expose a `Settings` class that owns the runtime-settable preferences and operations on them:

```typescript
export class Settings {
  constructor(private config: Config) {}
  get model(): string { ... }
  get effort(): EffortValue { ... }
  async pickModel(): Promise<void> { ... }
  async pickEffort(): Promise<void> { ... }
}
```

The command registry maps `/model` → `settings.pickModel()` and `/effort` → `settings.pickEffort()`. `Settings` doesn't know about commands — it exposes operations; the registry wires them to slash commands.

This also naturally holds a reference to `config`, so initial values come from `config.model` / `config.effort` without copying them into separate fields.

jasoncrawford:
## Summary

- Merges `src/agent/effort.ts` and `src/agent/model.ts` into a new `src/agent/settings.ts`
- Introduces a `Settings` class that owns runtime-settable preferences (`model` and `effort`) and exposes `pickModel()`/`pickEffort()` operations
- The command registry maps `/model` → `settings.pickModel()` and `/effort` → `settings.pickEffort()`; `Settings` has no knowledge of commands
- Adds `tests/settings.test.ts` covering the `Settings` class directly
- Updates all import sites (`index.ts`, `worker.ts`, test files) to use `settings.js`

## Test plan

- [ ] `/model` and `/effort` commands work interactively in the worker REPL

Closes #689


src/agent/settings.ts
```diff
@@ -113,3 +183,37 @@ export async function handleModelCommand(
...
+    this._model = await handleModelCommand(args, this._model, pickFn, fetchModelsFn, print);
+  }
+
+  async pickEffort(
+    args: string,
+    pickFn: PickFn,
+    print: (msg: string) => void,
+  ): Promise<void> {
+    this._effort = await handleEffortCommand(args, this._effort, pickFn, print);
+  }
```
jasoncrawford, line 195: instead defining handle*Command separately and then invoking them here, they should just be inlined here

tests/effort.test.ts
jasoncrawford, line 1: let's also combine the effort and model test files into a single settings test?

jasoncrawford (commented):
Both addressed:
- Inlined `handleModelCommand` and `handleEffortCommand` directly into `Settings.pickModel()` / `Settings.pickEffort()` — the standalone helpers are gone
- Merged `effort.test.ts` and `model.test.ts` into `settings.test.ts`, all tests now exercise the `Settings` class

src/agent/index.ts
```diff
@@ -304,26 +303,24 @@ export async function main(
...
-      const newModel = await handleModelCommand(
+      await settings.pickModel(
         args,
-        statusBar.model,
         (opts, idx) => pick(opts, { currentIdx: idx, escapable: true }),
         fetchModelsFn,
         display.print,
       );
-      statusBar.update({ model: newModel });
+      statusBar.update({ model: settings.model });
```
jasoncrawford: It feels to me like Settings should be an event emitter, and emit changed events, which the status bar should listen to?

src/agent/index.ts
```diff
@@ -233,7 +231,9 @@ export async function main(
   initialModel?: string,
   initialEffort?: EffortValue,
 ): Promise<void> {
+  const settings = new Settings({ model: initialModel, effort: initialEffort });
   statusBar.update({ model: initialModel, effort: initialEffort });
+  settings.on("change", () => statusBar.update({ model: settings.model, effort: settings.effort }));
```
jasoncrawford: I think the status bar should be responsible for doing its own listening.

suggestion:
- give status bar a settings object on constructor
- move creation and initialization of the status bar to right here, instead of later in this file
- have status bar do this on-settings-change call

PR #695, opened 2026-04-16
feat: add foreman_error message to wire protocol

### LINKED ISSUE #694: Add error message to foreman→worker wire protocol
The foreman has no way to communicate errors back to the worker. If something catastrophic happens (DB write failure, unhandled exception) during hello handling or task assignment, the worker gets a dropped connection with no explanation and retries indefinitely.

An HTTP/API server would return a 5xx response in this situation. Our WebSocket protocol has no equivalent.

## Proposed addition to wire protocol

Add a `foreman_error` message type to the foreman→worker wire protocol in `shared/wire.ts`:

```typescript
{ type: "foreman_error", message: string, fatal: boolean }
```

- `fatal: true` — the worker should print the error clearly and drop back to interactive REPL mode (not exit the process — the user may want to investigate or retry manually)
- `fatal: false` — informational; worker prints it and continues

## Where to send it

**On hello:** if a catastrophic error occurs during `handleBusyHello` or `handleIdleHello` (e.g. a DB write fails, an unexpected exception is thrown), send `foreman_error` before closing the connection instead of just dropping it.

**Generally:** any unhandled exception in a foreman message handler that currently gets swallowed or logged should send a `foreman_error` to the worker if one is connected for that task. This is the complement to issue #670 (error handling) — #670 is about not swallowing errors internally; this gives the protocol a channel to surface them to the other side.

## Worker behavior

On receiving `foreman_error`:
- Print the message clearly (bold red or similar)
- If `fatal`: exit worker mode and return to the REPL prompt
- If not `fatal`: print and continue

jasoncrawford:
## Summary

- Adds `{ type: "foreman_error", message: string, fatal: boolean }` to the `ForemanMessage` wire type in `shared/wire.ts`
- Foreman sends this instead of silently dropping the connection when catastrophic errors occur during hello handling or message processing
- Worker displays it in bold red; if `fatal`, stops reconnecting and returns to interactive REPL mode (no process exit)

## Changes

**Wire protocol** (`shared/wire.ts`): New `foreman_error` variant on `ForemanMessage`.

**Foreman** (`src/foreman/controllers/wss.ts`):
- `sendError(ws, message, fatal)` — sends `foreman_error` directly to a socket
- `handleIdleHello`: sends non-fatal error when task revert fails (previously just logged and returned)
- `handleBusyHello`: wrapped in try/catch; sends non-fatal error on unexpected exceptions
- Outer message handler catch: sends non-fatal error so workers know something went wrong

**Worker** (`src/agent/worker.ts`, `src/agent/index.ts`):
- Non-fatal: prints and continues (handles display via `FOREMAN_MESSAGE_FMT`)
- Fatal: aborts running query, stops reconnecting (`stopped` flag), signals `main()` via new `WS_FATAL` sentinel → drops to interactive REPL prompt

**Display** (`src/agent/display.ts`): `foreman_error` always prints (not verbose-only) in bold red with `[foreman error]` prefix.

**Docs** (`docs/type-system.md`): Updated `ForemanMessage` example.

## Test plan

- [x] `tests/repl.foreman-message.test.ts` — foreman_error always prints, not verbose-gated, includes error prefix
- [x] `tests/worker.test.ts` — non-fatal continues, fatal stops reconnecting + resolves WS_FATAL + aborts query
- [x] `tests/foreman.hello-handlers.test.ts` — idle hello revert failure sends foreman_error; busy hello assign failure sends foreman_error
- [x] All 1445 unit tests passing; type check clean; lint clean

Closes #694

### DIFF
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -10,7 +10,7 @@ import { setThinkOutLoud, setVerbose, verbose } from "./display.js";
 import { StatusBar, statusBar, initStatusBar } from "./status-bar.js";
 import { ask, pick, pickMultiple, pickQuestion } from "./input.js";
 import type { PickQuestionResult } from "./input.js";
-import { WorkerSession, registerWorkerCommands, startWorkerMode, generateAgentId } from "./worker.js";
+import { WorkerSession, registerWorkerCommands, startWorkerMode, generateAgentId, WS_FATAL } from "./worker.js";
 import type { RunQuery, WorkerModeConfig } from "./worker.js";
 import { loadConfig } from "../config.js";
 import { Workspace, confirmIfUnsafe, registerWorkspaceCommands } from "./workspace.js";
@@ -364,6 +364,13 @@ export async function main(
     // Ignore the internal abort sentinel (fired when wsAbort resolves at the
     // same tick as the ask() call; never a user action).
     if (input === "__abort__") continue;
+
+    // WS_FATAL: a fatal foreman_error was received — drop back to interactive REPL.
+    // The session has already stopped reconnecting; just show the prompt and continue.
+    if (input === WS_FATAL) {
+      showPrompt = true;
+      continue;
+    }
 
     // WS_PROMPT: a task prompt or debounced event prompt is ready. Hide the
     // prompt, drain all queued prompts through runQueryFn, then show the prompt
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -10,7 +10,7 @@ import { setThinkOutLoud, setVerbose, verbose } from "./display.js";
 import { StatusBar, statusBar, initStatusBar } from "./status-bar.js";
 import { ask, pick, pickMultiple, pickQuestion } from "./input.js";
 import type { PickQuestionResult } from "./input.js";
-import { WorkerSession, registerWorkerCommands, startWorkerMode, generateAgentId, WS_FATAL } from "./worker.js";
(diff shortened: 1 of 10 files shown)


src/agent/index.ts
```diff
@@ -365,6 +365,13 @@ export async function main(
     // same tick as the ask() call; never a user action).
     if (input === "__abort__") continue;
 
+    // WS_FATAL: a fatal foreman_error was received — drop back to interactive REPL.
+    // The session has already stopped reconnecting; just show the prompt and continue.
+    if (input === WS_FATAL) {
```
jasoncrawford: shouldn't this use Worker.isFatalSignal() and not import the private WS_FATAL sentinel?

src/agent/worker.ts
```diff
@@ -131,6 +131,8 @@ export type WorkerModeConfig = {
 
 // Sentinel: a prompt is ready for main() to execute
 export const WS_PROMPT = "__ws_prompt__";
+// Sentinel: a fatal foreman error was received; main() should drop back to interactive REPL
+export const WS_FATAL = "__ws_fatal__";
```
jasoncrawford: I think these sentinels should be private?

jasoncrawford (commented):
Good catch on both — fixed in 2ab89a6. `WS_PROMPT` and `WS_FATAL` are now unexported module-private constants, and `index.ts` uses `WorkerSession.isFatalSignal()` instead of importing the sentinel directly. Tests updated to go through the static methods too.

PR #697, opened 2026-04-16
refactor: read config values directly instead of copying into local state

### LINKED ISSUE #691: Read config values directly instead of copying them into local state
Several places copy config values into local variables, class fields, or intermediate typed objects rather than reading them from the config object directly. The root cause is that `config.ts` exports only an async `loadConfig()` function — there is no singleton any module can import. So modules that need config values at render time (like `display.ts` needing `verbose`) have no choice but to store their own copy.

## Step 1: Export a config singleton

Add a `getConfig()` function to `config.ts` that returns the loaded config, and have `loadConfig()` store the result before returning it:

```typescript
let _config: BrunelConfig | null = null;

export function getConfig(): BrunelConfig {
  if (!_config) throw new Error("Config not initialized — call loadConfig() first");
  return _config;
}

export async function loadConfig(...): Promise<BrunelConfig> {
  // ... existing logic ...
  _config = result;
  return result;
}
```

Entry points already call `loadConfig()` before anything else, so `getConfig()` will always be safe by the time any module uses it.

## Step 2: Clean up the copies

Once the singleton exists, remove these patterns:

**`verbose` and `thinkOutLoud` in `display.ts`** — module-level `let` variables set via `setVerbose(config.verbose)` at startup. Replace with `getConfig().verbose` at point of use. Delete the variables and setters.

**`WorkerModeConfig` type in `worker.ts`** — `index.ts` constructs a subset of config into a new typed object to pass to `startWorkerMode()`. This intermediate type exists only to shuttle config fields between two files. `startWorkerMode()` can call `getConfig()` directly and the type can be deleted.

**`ForemanWss` instance variables** — the constructor copies `config.githubRepo`, `config.githubToken`, `config.githubApiUrl`, `config.taskLabel` into `this.repo`, `this.token`, etc. The class already receives config; it should call `getConfig()` (or hold `this.config`) and read fields directly.

jasoncrawford:
## Summary

- Export `getConfig()` singleton from `config.ts` so modules read config fields directly
- Remove `verbose`/`thinkOutLoud` module-level vars from `display.ts`; `setVerbose`/`setThinkOutLoud` now mutate the config singleton
- Remove `WorkerModeConfig` type from `worker.ts`; `startWorkerMode()` calls `getConfig()` directly
- Remove copied instance variables from `ForemanWss`; hold `this.config` and read fields as needed
- Fix pre-existing circular import: `status-bar.ts` now imports `getConfig` from `config.ts` instead of `verbose` from `display.ts`

## Test plan

- [ ] `npm test` — all unit tests pass (1438 tests)
- [ ] `npx tsc --noEmit` — type check clean
- [ ] `tests/setup.ts` calls `loadDefaultConfig()` so `getConfig()` is initialized before any test runs `setVerbose()`

Closes #691


src/agent/display.ts
```diff
@@ -15,18 +16,13 @@ export const VERBOSE_PREFIX_LEN = 9;
...
 
-export let verbose = false;
-export function setVerbose(v: boolean) { verbose = v; }
-
-// ── Think-out-loud flag ───────────────────────────────────────────────────────
-
-export let thinkOutLoud = false;
-export function setThinkOutLoud(v: boolean) { thinkOutLoud = v; }
+export function setVerbose(v: boolean) { getConfig().verbose = v; }
+export function setThinkOutLoud(v: boolean) { getConfig().thinkOutLoud = v; }
```
jasoncrawford: Wait, why do we need setters for these? They are never set. They are constants loaded from config. Delete these setters.

jasoncrawford (commented):
Good catch — you're right. They're only called from tests at this point, and they're just `getConfig().verbose = v` wrappers. I'll remove them and have the tests mutate the config directly.

src/foreman/controllers/wss.ts
```diff
@@ -499,21 +493,21 @@ export class ForemanWss {
...
         const enqueued = await this.taskManager.handleIssueLabeledEvent(
           issueNumber,
-          this.repo,
+          this.config.githubRepo,
           String(issue.title ?? ""),
           String(issue.body ?? ""),
           labels,
           String(issue.state ?? "open"),
-          { repo: this.repo, token: this.token, apiUrl: this.githubApiUrl },
+          { repo: this.config.githubRepo, token: this.config.githubToken, apiUrl: this.config.githubApiUrl },
```
jasoncrawford: can't the taskManager also access config? so it should be able to get these things directly, and they don't need to be passed as arguments

src/foreman/controllers/wss.ts
```diff
@@ -567,7 +561,7 @@ export class ForemanWss {
         this.taskManager.handleIssueBodyEditedEvent(
           issueNumber,
           String(issue.body ?? ""),
-          { repo: this.repo, token: this.token, apiUrl: this.githubApiUrl },
+          { repo: this.config.githubRepo, token: this.config.githubToken, apiUrl: this.config.githubApiUrl },
```
jasoncrawford: again, taskManager should access config directly

jasoncrawford (commented):
Good point — I'll have TaskManager import getConfig() directly and drop these arguments.

src/foreman/models/task-manager.ts
```diff
@@ -217,12 +218,12 @@ export class TaskManager extends EventEmitter {
   async fetchAndLoadDeps(
     issueNumber: number,
     body: string,
-    config: { repo: string; token: string; apiUrl?: string },
   ): Promise<void> {
-    const blockers = await Task.fetchBlockers(issueNumber, body, config);
+    const { githubRepo: repo, githubToken: token, githubApiUrl: apiUrl } = getConfig();
+    const blockers = await Task.fetchBlockers(issueNumber, body, { repo, token, apiUrl });
```
jasoncrawford: Can the task model also get these things off the config object?

src/foreman/models/task-manager.ts
```diff
@@ -217,12 +218,12 @@ export class TaskManager extends EventEmitter {
...
     this.setBlockers(issueNumber, blockers);
     if (blockers.length > 0) {
-      const states = await fetchIssueStates(blockers, { repo: config.repo, token: config.token });
+      const states = await fetchIssueStates(blockers, { repo, token });
```
jasoncrawford: why are we passing more config options into this function?

src/foreman/models/task-manager.ts
```diff
@@ -334,10 +332,8 @@ export class TaskManager extends EventEmitter {
 
   /** Fetch brunel:ready issues from GitHub and load deps.
    *  Called at startup after loadActiveTasksFromDb. */
-  async loadIssuesFromGithub(
-    config: { githubRepo: string; githubToken: string; taskLabel: string; githubApiUrl?: string },
-  ): Promise<void> {
-    await loadIssuesToQueue(this, config);
+  async loadIssuesFromGithub(): Promise<void> {
+    await loadIssuesToQueue(this, getConfig());
```
jasoncrawford: why can't loadIssuesToQueue look up the config itself?

jasoncrawford (commented):
Right, the pattern should go all the way down. I'll have `fetchNativeBlockers`, `Task.fetchBlockers`, `fetchIssueStates`, and `loadIssuesToQueue` all call `getConfig()` directly.

PR #700, opened 2026-04-16
fix: persist foreman_error messages to activity log

### LINKED ISSUE #699: sendError bypasses ForemanMessage.log() — foreman errors not persisted to activity log
## Problem

`sendError` in `src/foreman/controllers/wss.ts` sends `foreman_error` messages directly over the WebSocket without calling `ForemanMessage.log()`. This means these errors are never written to the `foreman_messages` DB table and never appear in the activity log.

These are some of the most important messages to have a record of — a fatal error that causes a worker to drop back to REPL is exactly the kind of event you'd want to see in the dashboard.

## What to do

Call `ForemanMessage.log()` (or equivalent) inside `sendError` before or after sending the wire message, so foreman errors are persisted. The activity log query in `activity-log.ts` already merges `foreman_messages` by timestamp, so logged errors should appear automatically.

Note: `sendError` is called in contexts where a `Worker` model object may not yet exist (e.g. during `handleIdleHello` before the worker is registered). `ForemanMessage.log()` may need a nullable worker ID, or a separate logging path for pre-registration errors.

jasoncrawford:
## Summary

- `sendError()` in `wss.ts` now calls `ForemanMessage.log()` and `broadcastMessageEvent()` alongside sending the wire message, so foreman errors appear in the activity log and admin dashboard
- `workerId` and `taskId` are threaded through all three `sendError` call sites; the `handleIdleHello` revert-failure path also passes the prior task's ID
- `ForemanMessage.buildSummary()` gains a `foreman_error` branch that renders the error message and marks fatal errors with `(fatal)`

## Test plan

- [ ] New tests in `foreman.hello-handlers.test.ts` verify that `ForemanMessage.log()` is called with the correct `direction`, `workerId`, `taskId`, and `payload` when `sendError` fires from both `handleIdleHello` and `handleBusyHello` error paths
- [ ] New tests verify `ForemanMessage.buildSummary()` renders `foreman_error` summaries correctly (non-fatal vs fatal)
- [ ] All 1456 existing unit tests continue to pass; type check and lint are clean

Closes #699

### DIFF
--- a/src/foreman/controllers/wss.ts
+++ b/src/foreman/controllers/wss.ts
@@ -254,11 +254,13 @@ export class ForemanWss {
 
   sendMsg(worker: Worker, msg: Wire.ForemanMessage, logTaskId?: string): void {
     const taskId = logTaskId ?? (("taskId" in msg ? msg.taskId : null) ?? null);
-    const workerId = worker.workerId;
     worker.send(msg);
-    const msgPayload = msg as unknown as Record<string, unknown>;
-    void ForemanMessage.log({ direction: "sent", workerId, taskId, msgType: msg.type, payload: msgPayload });
-    this.broadcastMessageEvent({ direction: "sent", workerId, taskId, msgType: msg.type, payload: msgPayload });
+    this.logAndBroadcastSent(worker.workerId, taskId, msg.type, msg as unknown as Record<string, unknown>);
+  }
+
+  private logAndBroadcastSent(workerId: string | null, taskId: string | null, msgType: string, payload: Record<string, unknown>): void {
+    void ForemanMessage.log({ direction: "sent", workerId, taskId, msgType, payload });
+    this.broadcastMessageEvent({ direction: "sent", workerId, taskId, msgType, payload });
   }
 
   private broadcastMessageEvent(data: { direction: string; workerId: string | null; taskId: string | null; msgType: string; payload?: Record<string, unknown> }): void {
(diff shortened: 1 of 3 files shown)


src/foreman/controllers/wss.ts
```diff
@@ -296,11 +296,16 @@ export class ForemanWss {
...
+  private sendError(ws: WebSocket, message: string, fatal: boolean, workerId: string | null, taskId: string | null = null): void {
+    const payload: Wire.ForemanMessage = { type: "foreman_error", message, fatal };
     if (ws.readyState === WebSocket.OPEN) {
-      ws.send(JSON.stringify({ type: "foreman_error", message, fatal } satisfies Wire.ForemanMessage));
+      ws.send(JSON.stringify(payload));
     }
+    const msgPayload = payload as unknown as Record<string, unknown>;
+    void ForemanMessage.log({ direction: "sent", workerId, taskId, msgType: payload.type, payload: msgPayload });
+    this.broadcastMessageEvent({ direction: "sent", workerId, taskId, msgType: payload.type, payload: msgPayload });
   }
```
jasoncrawford, line 303: I feel that this and sendMsg should go through some common function that does these core things, rather than the two methods duplicating this logic

jasoncrawford:
Good call — extracted `logAndBroadcastSent` as a private helper that both `sendMsg` and `sendError` now call. The duplication is gone.

PR #709, opened 2026-04-17
feat: apply MVC structure to src/agent/ (models, views, controllers)

### LINKED ISSUE #706: Apply MVC structure to agent: subdirectories + layer separation
The agent (`src/agent/`) has no directory structure — everything is flat — and no clear MVC separation. The foreman already has `models/`, `controllers/`, and `clients/` subdirectories. The agent should follow the same pattern.

## Proposed directory structure

```
src/agent/
  models/
    workspace.ts      — git/npm workspace state and operations
    settings.ts       — user preferences (model, effort)
    query-stats.ts    — per-query token/turn tracking
  views/
    display.ts        — Display class: renders output, receives config in constructor
    status-bar.ts     — StatusBar class: reactive view-model, subscribes to model events
    input.ts          — core input primitives: ask(), pick(), multiline
  controllers/
    worker.ts         — WorkerSession: WebSocket/foreman protocol controller
    command-registry.ts — command registration and dispatch
  index.ts            — composition root + REPL controller (stays at top level as entry point)
  worker-prompts.ts   — prompt templates (already renamed from templates.ts)
```

## Layer behavior changes

**`Workspace` → model.** Remove all `display.print()` calls. Emit events instead (`"created"`, `"reset"`, `"branch-changed"`, etc.). The controller that owns the workspace subscribes and routes to display. Models should have no knowledge of the display layer.

**`Display` → view class.** Convert from a bag of module-level functions to a `Display` class. Receives config in its constructor; methods use `this.config.verbose` instead of `getConfig().verbose`. Export a singleton initialized at startup from the entry point.

**`WorkerSession` → controller.** Receives `display` injected in its constructor. Calls `display.print()` directly — that is appropriate for a controller.

## Connections to other issues

Several related issues should land first or alongside this one:
- #677 — command dispatch/autocomplete into CommandRegistry (affects what stays in `input.ts`)
- #683 — merge `effort.ts` + `model.ts` into `settings.ts` (already done, determines `models/settings.ts`)
- #674 / PR #681 — StatusBar as a class (affects `views/status-bar.ts`)
- #691 — config singleton / `getConfig()` (affects how `Display` accesses config)

## Import path updates

Moving files into subdirectories requires updating all import paths that reference them. `grep -r "from.*agent/" src/` will find external references; internal cross-references within `src/agent/` will also need updating.

jasoncrawford:
## Summary

- **Models** (`src/agent/models/`): `workspace.ts`, `settings.ts`, `query-stats.ts` — pure state, no UI imports; `Workspace` extends `EventEmitter` and emits events instead of calling `print()` directly
- **Views** (`src/agent/views/`): `display.ts` (new `Display` class with `BrunelConfig` injected; standalone delegate functions `print`/`printMessage`/etc. for testability), `status-bar.ts`, `input.ts`
- **Controllers** (`src/agent/controllers/`): `worker.ts`, `workspace-commands.ts` (extracted from workspace.ts), `command-registry.ts`
- `index.ts` calls `initDisplay(new Display(config))` at startup; `tests/setup.ts` does the same for tests
- All old flat `src/agent/*.ts` files removed; all test imports updated

## Test plan
- [ ] All 1482 unit tests pass (only 3 DB tests fail pre-existing, require Supabase)
- [ ] `npx tsc --noEmit` passes cleanly
- [ ] CI passes on this PR

Closes #706


src/agent/controllers/worker.ts
```diff
@@ -4,16 +4,16 @@ import { promisify } from "node:util";
...
 import { WebSocket } from "ws";
-import * as display from "./display.js";
-import { StatusBar, statusBar } from "./status-bar.js";
-import { buildInitialPrompt, buildEventPrompt } from "./worker-prompts.js";
-import type { EffortValue } from "./settings.js";
-import * as Wire from "../../shared/wire.js";
-import { Workspace, confirmIfUnsafe } from "./workspace.js";
-import { fmtError } from "../utils.js";
-import { getConfig } from "../config.js";
+import { c, print, printForemanMessage } from "../views/display.js";
```
jasoncrawford: I don't think this should be importing methods directly from the display module. It should rely on the display object it receives.

Well, maybe `c` is ok to import since that's going to be pretty self-contained and it helps to have a nice shorthand for it. But I think print methods should always be called on the display object.

src/agent/controllers/worker.ts
```diff
@@ -707,8 +729,8 @@ export async function startWorkerMode(): Promise<{
   };
 
   const workerDisplay: WorkerDisplay = {
```
jasoncrawford: shouldn't WorkerDisplay just receive a display object, rather than loose functions?

src/agent/models/query-stats.ts
jasoncrawford, line 1: Since this is a model it shouldn't depend on views. (1) pickWorkingVerb doesn't have to be a view method, it could just be moved into this file if nothing else uses it? Or it could be injected on the constructor. (2) getStatusText could be moved outside this file, something else could do the formatting. Or fmtStats could be moved in here, if it isn't used anywhere else.

src/agent/models/settings.ts
jasoncrawford, line 1: Most of this is controller logic, not model. I think we should have a small Settings model here, and the pick action should move into a new SettingsController in the controllers directory.

src/agent/controllers/workspace-controller.ts
jasoncrawford, line 1: maybe call this the workspace controller?

src/agent/views/display.ts
```diff
@@ -648,106 +652,181 @@ export const FOREMAN_MESSAGE_FMT: FmtTable = {
...
-function printLine(line: string): void {
-  if (getConfig().verbose) {
-    const ts = `\x1b[90m${fmtTime()} \x1b[39m`;
-    const parts = line.split("\n");
-    const openColor = (line.match(/^(\x1b\[[0-9;]*m)+/) ?? [""])[0];
-    console.log(parts.map((p, i) => ts + (i > 0 ? openColor : "") + p).join("\n"));
-  } else {
-    console.log(line);
+export const toolUseNames = new Map<string, string>();
+export const toolUseInputs = new Map<string, Record<string, unknown>>();
```
jasoncrawford: why are these exported?

src/agent/views/display.ts
```diff
@@ -648,106 +652,181 @@ export const FOREMAN_MESSAGE_FMT: FmtTable = {
...
 }
 
-export function printMessage(msg: unknown) {
-  if (msg === null || typeof msg !== "object") return;
-  const m = msg as Record<string, unknown>;
+// ── Singleton ─────────────────────────────────────────────────────────────────
 
-  if (m.parent_tool_use_id != null) return;
+/** Shared singleton. Call initDisplay() at startup before first use. */
+export let display: Display = undefined!;
```
jasoncrawford: I don't think we should have a shared singleton that lives in this class. The main app should create the display object and pass it to the controllers

src/agent/views/display.ts
```diff
@@ -648,106 +652,181 @@ export const FOREMAN_MESSAGE_FMT: FmtTable = {
...
-    if (!content.length) { print(resolve(MESSAGE_FMT, "_empty", m)); return; }
-    for (const b of content) printBlock(b, role, m);
-    return;
-  }
+// ── Standalone printing delegates ─────────────────────────────────────────────
+//
+// These delegate to the singleton Display instance. They exist so that callers
+// (including tests) can import and call print/printMessage/etc. directly without
+// needing to import the singleton. The Display must be initialized via
+// initDisplay() before these are called.
```
jasoncrawford: This defeats a lot of the purpose of this reorg. We should not be exporting standalone functions. Everything should obtain a display object, either creating one for itself or receiving one on its constructor, and call methods on that.

src/agent/index.ts
```diff
@@ -440,6 +442,7 @@ export async function main(
 
 if (process.argv[1] === fileURLToPath(import.meta.url)) {
   const config = await loadConfig(process.argv);
+  initDisplay(new Display(config));
```
jasoncrawford: instead of initDisplay, this should pass the display object to the controllers it instantiates.

src/agent/controllers/command-controller.ts
jasoncrawford, line 1: what if we rename this command-controller.ts and the object CommandController? just to drive home the MVC architectural intent

src/agent/controllers/worker-controller.ts
jasoncrawford, line 1: rename this file/object to be "worker controller"?

jasoncrawford (commented):
See detailed comments below. High-level: the intent here is to achieve a true MVC architecture for the agent. Specifically, as noted below:

1. The controllers should be named as such.
2. The models should not have view dependencies (in particular, I think settings needs to be split into model and controller).
3. The display object should not be a singleton but should be injected into the controllers.

jasoncrawford (commented):
Good architectural feedback — I agree with all the points. Here's what I'll do:

1. **Remove display singleton + standalone delegates** — `Display` will be constructed in `index.ts` and injected into every controller that needs it. No `initDisplay` or module-level `print` export.

2. **`WorkerSession` display injection** — I'll define a `WorkerDisplay` interface in `views/display.ts` (not in the worker controller) that `Display` satisfies structurally. `WorkerSession` will be typed against that interface so tests can still pass minimal mocks without `as unknown as Display`, while production code passes the real instance.

3. **Settings split** — Small `Settings` model (just `model`/`effort` state + EventEmitter); new `SettingsController` in controllers/ with `pickModel`/`pickEffort`.

4. **`QueryStats`** — Move `WORKING_VERBS`/`pickWorkingVerb` into `query-stats.ts` (reviewer's option 1). Inline the tiny format helpers (`fmtDuration`, `fmtNum`) locally so there are zero view imports.

5. **`toolUseNames`/`toolUseInputs`** — Private instance state on `Display`, exposed as a public getter for test cleanup.

6. **Renames** — `command-registry.ts → command-controller.ts` / `CommandController`, `workspace-commands.ts → workspace-controller.ts`, `worker.ts → worker-controller.ts`.

7. **`runQuery` + `main`** — Both receive `display: Display` as a parameter so there is no ambient global dependency.

jasoncrawford (commented):
All five points are now implemented — just pushed commit 9155905.

**What changed:**

1. **Display singleton removed** — `initDisplay` and standalone delegates (`print`, `printBlock`, `printMessage`, `printForemanMessage`) are gone from `views/display.ts`. `Display` is constructed once in `index.ts` and passed as a parameter to `runQuery(display, ...)`, `main(..., display, ...)`, `startWorkerMode(display)`, `registerWorkerCommands(session, registry, display)`, `registerWorkspaceCommands(workspace, registry, display)`, and `confirmTaskQuit(info, display, pickFn?)`.

2. **`WorkerDisplay` interface** — Defined in `views/display.ts` as `{ print(line: string | null): void; printForemanMessage(msg: Wire.ForemanMessage): void }`, satisfied structurally by `Display`. Tests pass minimal plain-object mocks; production code passes the real `Display` instance. No `as unknown as` needed.

3. **Settings split** — `Settings` is now a pure model (EventEmitter, `_setModel`/`_setEffort`, no view imports). New `controllers/settings-controller.ts` exports `SettingsController` which wraps `Settings` with an injected `WorkerDisplay` and owns `pickModel()`/`pickEffort()`.

4. **`query-stats.ts` clean** — `WORKING_VERBS` and `pickWorkingVerb` moved from `views/display.ts` into `models/query-stats.ts`. Tiny format helpers (`fmtDuration`, `fmtNum`, `fmtStats`) inlined locally — no view imports. `QueryStats` accepts an optional `pickVerb` parameter for test injection (spying on intra-module calls via `vi.spyOn` doesn't intercept direct calls in ESM).

5. **`toolUseNames`/`toolUseInputs`** — Private instance fields on `Display`; public `get toolUseNames()` getter for test cleanup.

6. **Renames** — `command-registry.ts` → `command-controller.ts` / `CommandRegistry` → `CommandController`; `workspace-commands.ts` → `workspace-controller.ts`; `worker.ts` → `worker-controller.ts`.

All 1482 non-DB tests pass.

src/agent/controllers/command-controller.ts
jasoncrawford, line 1: Now that I look at this again, I think it might make sense to have two objects in here: a CommandRegistry that has the `register` and `scoped` methods, and a CommandController that parses commands and invokes them… I'm not exactly sure where to draw the boundary between them, but I think other controllers should register their commands with a registry instead of with another controller.

src/agent/controllers/settings-controller.ts
```diff
@@ -0,0 +1,155 @@
+import { c } from "../views/display.js";
+import type { WorkerDisplay } from "../views/display.js";
+import type { PickResult } from "../views/input.js";
+import {
+  Settings,
+  EFFORT_LEVELS,
+  findModel,
+  getCachedModels,
+  setCachedModels,
```
jasoncrawford: Why does it have to import so many things? Why can't these properties/methods be available on the settings object or class?

src/agent/controllers/worker-controller.ts
```diff
@@ -140,16 +136,17 @@ export type TaskQuitInfo = {
  */
 export async function confirmTaskQuit(
   info: TaskQuitInfo,
+  display: WorkerDisplay,
```
jasoncrawford: why can't this just be Display? why do we need a separate WorkerDisplay type?

jasoncrawford: also why does this even need to be an argument? shouldn't the controller have a `this.display`?

src/agent/models/query-stats.ts
```diff
@@ -1,5 +1,62 @@
 import { EventEmitter } from "node:events";
-import * as display from "./display.js";
+
+// ── Working verbs ─────────────────────────────────────────────────────────────
+
+export const WORKING_VERBS = [
```
jasoncrawford: does this need to be exported anymore? could it be private?

src/agent/models/query-stats.ts
```diff
@@ -27,10 +84,10 @@ export class QueryStats extends EventEmitter {
   private readonly _startTime: number;
   private readonly _workingVerb: string;
 
-  constructor(startTime = Date.now()) {
+  constructor(startTime = Date.now(), pickVerb: () => string = pickWorkingVerb) {
```
jasoncrawford: why is this on the constructor if it is defined in this file? just make it a private method in this file and call it.

src/agent/models/settings.ts
jasoncrawford, line 1: Everything in this file should go on the object or class, and the class should be the only export.

src/agent/workspace.ts
```diff
@@ -158,11 +156,9 @@ export class Workspace {
       if (fs.existsSync(lockPath)) {
         const pid = parseInt(fs.readFileSync(lockPath, "utf8").trim(), 10);
         if (isProcessAlive(pid)) {
-          if (getConfig().verbose) display.print(display.c.sageGreen(`[workspace] Skipping active workspace ${dir} (pid ${pid})`));
           continue;
         }
       }
-      display.print(display.c.sageGreen(`[workspace] Removing orphaned workspace ${dir}`));
```
jasoncrawford, line 165: these things don't get events emitted? did we lose the messages printed for them?

src/agent/views/display.ts
```diff
@@ -648,133 +652,162 @@ export const FOREMAN_MESSAGE_FMT: FmtTable = {
...
-    console.log(parts.map((p, i) => ts + (i > 0 ? openColor : "") + p).join("\n"));
-  } else {
-    console.log(line);
-  }
+/**
+ * Minimal display interface required by controllers (WorkerSession,
+ * SettingsController, workspace/worker command registrations).
+ * Satisfied by the Display class; tests can pass lightweight stubs.
+ */
+export interface WorkerDisplay {
```
jasoncrawford: I don't see why we need WorkerDisplay, I would just drop it. If we really want it for some reason, I think it should be defined in worker-controller

jasoncrawford:
Good feedback — implemented all points in commit 1ea8b88. Here's what changed:

**CommandRegistry / CommandController split** — `CommandRegistry` is now the base class with just `register`, `scoped`, `lookup`, `listAll`, `execute`, and `_reset`. `CommandController extends CommandRegistry` and adds `dispatch`, `parseSlashCommand`, `listCommandNames`, `listCommands`, and `suggest`. Registration functions (`registerWorkspaceCommands`, `registerWorkerCommands`) are now typed to accept `CommandRegistry`. Since `CommandController` extends it, all existing `new CommandController()` calls work unchanged, and all tests continue to pass without modification.

**Settings static members** — `Settings` is now the only value export from `settings.ts`. `EFFORT_LEVELS`, `VALID_EFFORT_VALUES`, `getCachedModels`, `setCachedModels`, `_resetCachedModels`, and `findModel` are all static members of the class. `SettingsController` now only imports `Settings` (plus a few types).

**WorkerDisplay moved** — Removed from `views/display.ts` and moved to `controllers/worker-controller.ts` where it belongs. `workspace-controller.ts` and `settings-controller.ts` import it from there.

**confirmTaskQuit as method** — Removed the standalone function. It's now `WorkerSession.confirmTaskQuit(info, pickFn?)` which uses `this.display`. Updated `startWorkerMode` accordingly.

**WORKING_VERBS / pickWorkingVerb unexported** — Removed `export` from both; removed the `pickVerb` constructor injection from `QueryStats`. Deleted `tests/repl.working-verb.test.ts` (was testing unexported implementation details).

**workspace.ts prune messages** — No messages were lost. `Workspace.prune()` is called only by the `/workspace:prune` command handler in `workspace-controller.ts`, which prints per-directory and summary messages via `display`. The `destroy()` event is still subscribed in `startWorkerMode()` and prints `[workspace] Destroying ${dir}`.

src/agent/controllers/command-controller.ts
```diff
@@ -336,6 +336,18 @@ export class CommandRegistry {
...
+    this._root._entries.clear();
+  }
+}
+
+/**
+ * Full command controller: extends CommandRegistry with dispatch, parse, and
+ * list methods. Use this for the top-level registry; pass CommandRegistry to
+ * functions that only need to register commands.
+ */
+export class CommandController extends CommandRegistry {
```
jasoncrawford: No, a CommandController is not a type of CommandRegistry. It should just be a different thing. It should have a command registry on it.

src/agent/models/workspace.ts
```diff
@@ -148,7 +147,6 @@ export class Workspace {
    */
   static async prune(workspaceDir: string): Promise<string[]> {
```
jasoncrawford: on prune, it looks like we lost the messages printed about this action? the print calls were removed, but they weren't replaced with events that the controller could catch and log.

src/agent/index.ts
```diff
@@ -282,16 +289,16 @@ export async function main(
 
   // Register all commands. All commands are present in both REPL and worker
   // modes; commands that require a foreman connection degrade gracefully.
-  const registry = new CommandRegistry();
-  registerWorkspaceCommands(workspace, registry.scoped("workspace"));
-  registerWorkerCommands(session, registry.scoped("worker"));
+  const registry = new CommandController();
```
jasoncrawford: this should be a new CommandRegistry, not controller

src/agent/index.ts
```diff
@@ -5,19 +5,21 @@ import fs from "fs";
 import { fileURLToPath } from "url";
 import { query } from "@anthropic-ai/claude-agent-sdk";
 import type { CanUseTool, PermissionMode, PermissionResult } from "@anthropic-ai/claude-agent-sdk";
-import * as display from "./display.js";
-import { StatusBar, statusBar, initStatusBar } from "./status-bar.js";
-import { ask, pick, pickMultiple, pickQuestion } from "./input.js";
-import type { PickQuestionResult } from "./input.js";
-import { WorkerSession, registerWorkerCommands, startWorkerMode, generateAgentId, confirmTaskQuit } from "./worker.js";
-import type { RunQuery } from "./worker.js";
+import { Display, c, hr, clearBreak, fmtArgs } from "./views/display.js";
```
jasoncrawford: clearBreak and fmtArgs should not be imported, invoke them on the display object

src/agent/index.ts
```diff
@@ -234,16 +240,17 @@ export async function main(
   initStatusBar(new StatusBar({ agentId: generateAgentId(), settings }));
 
   // Worker mode setup: create workspace, session, signal handlers.
-  const workerCtx = runWorkerMode ? await startWorkerMode() : undefined;
+  const workerCtx = runWorkerMode ? await startWorkerMode(display) : undefined;
   const session = workerCtx?.session;
 
   const fetchModelsFn = createFetchModelsFn(permConfig);
+  const settingsCtrl = new SettingsController(settings, display);
```
jasoncrawford: don't abbreviate like this, call it `settingsController`

jasoncrawford:
All five points addressed in commit 5caafd2.

**CommandController composition** — `CommandController` no longer extends `CommandRegistry`. It has-a `CommandRegistry` via a private field, exposes it as `controller.registry` for cases where callers need to register (tests, etc.). `index.ts` now creates `new CommandRegistry()` for registration and `new CommandController(registry)` for dispatch.

**clearBreak / fmtArgs on display** — Added `readonly clearBreak = clearBreak` and `readonly fmtArgs = fmtArgs` as instance property delegates on `Display`. Removed the standalone imports from `index.ts`; callers now use `display.clearBreak()` and `display.fmtArgs(input)`.

**Workspace.prune() events** — Changed `prune()` from a static method to an instance method. It now emits `"prune-start"`, `"prune-remove"`, and `"prune-skip"` events (following the same pattern as `reset-start`, `clone-start`, etc.). `startWorkerMode()` subscribes to `"prune-start"` and `"prune-remove"` and prints the corresponding messages. No messages lost.

**new CommandRegistry()** — `index.ts` line 292 is now `const registry = new CommandRegistry()`.

**settingsController** — Renamed throughout `index.ts`.

PR #717, opened 2026-04-17
feat: extract AgentController from index.ts

### LINKED ISSUE #710: Extract AgentController from index.ts
After #706 (agent MVC structure) lands, `index.ts` still wears two hats: composition root and REPL controller. The controller responsibilities should move into an `AgentController` class, leaving `index.ts` as a thin entry point.

## What moves into AgentController

The query loop and everything that happens during a query:
- Running a query via the agent SDK
- Handling `handleAskUserQuestion` and `handleToolPermission` SDK callbacks
- Managing session ID across turns
- Printing the query result and reading the next prompt
- Displaying the model/effort banner at startup
- Permission mode display

## What stays in index.ts

Just the composition root:
- Load config
- Construct objects (`Display`, `StatusBar`, `Settings`, `AgentController`, `WorkerController`, etc.)
- Have each controller register its commands with the `CommandRegistry`
- Wire everything together and call `controller.start()`

`index.ts` should be readable in ~30 lines and serve as a table of contents for how the agent starts up.

Depends on #706 (agent MVC subdirectories) since `AgentController` would live in `src/agent/controllers/`.

jasoncrawford:
## Summary

- Extracts `AgentController` class into `src/agent/controllers/agent-controller.ts`, owning the Claude SDK query loop, tool permission callbacks, session ID management, startup banner, and command registration
- Reduces `index.ts` to a ~30-line composition root that constructs objects and calls `agentController.start()`
- Updates tests to use `AgentController` directly with `vi.spyOn(controller, 'runQuery')` instead of injecting a `runQueryFn` parameter

## Test plan

- [ ] All existing tests pass (`npm test` — 1479 tests, 49 files)
- [ ] Type check passes (`npx tsc --noEmit`)
- [ ] Lint passes (`npm run lint`)
- [ ] `index.ts` is now ~30 lines and readable as a table of contents for agent startup
- [ ] `AgentController.runQuery()` covers the same behavior as the former `runQuery()` function

Closes #710


jasoncrawford (commented):
This moves too much into AgentController / leaves too little in index.ts.

What should be in AgentController:
* running the query, given a prompt; runQuery()

What should stay in index/main:
* creating controllers
* creating the workspace
* starting worker mode
* the startup banner
* registering commands
* … basically, start()
* also, the core loop of getting input and then invoking either commands or a query

Think of it this way: index.ts is where we set up the object structure, and take input (either from the user typing, or from foreman messages). Then each input is routed to different controllers to be handled, depending on what type of input it is:

* command → invokes a handler method on the controller registered for that command/namespace
* user prompt → run a query via agent controller
* foreman message → invoke worker controller

So index.ts should do setup and routing; controllers should handle actions by invoking methods on models and views. This seems like the cleanest MVC architecture for the client.

jasoncrawford (commented):
Good point — the split was too coarse. I'll slim AgentController down to just `runQuery()` and put the setup/loop back in `main()` in `index.ts`. That keeps the routing clearly in `index.ts` (command → command handler, user prompt → `agentController.runQuery()`, foreman message → worker controller) while `AgentController` handles only the single action of executing one query turn against the SDK.

jasoncrawford (commented):
Done — pushed. The changes are in the latest commit on this branch.

PR #718, opened 2026-04-18
feat: add Input class to encapsulate display dependency

### LINKED ISSUE #715: Sort out responsibilities of input and display?
The awkwardness of `ask(display.statusBar, ...)` is a symptom of `input.ts` being a bag of functions rather than an object. The fix is to make `Input` a proper class with a `display` reference.

## Fix

Create an `Input` class in `src/agent/views/input.ts` (alongside `Display` in the views layer after #706):

```typescript
export class Input {
  constructor(private display: Display) {}
  async ask(prompt: string): Promise<string> { ... }
  async pick<T>(prompt: string, options: T[]): Promise<T> { ... }
  async pickMultiple<T>(prompt: string, options: T[]): Promise<T[]> { ... }
}
```

Internally, methods access `this.display.statusBar` for the redraw callbacks during input. Callers just do `input.ask("prompt")` — no statusBar threading required.

`Input` takes the full `Display` reference rather than a narrow interface — they are tightly coupled view siblings and the full reference is clean.

## Wiring

The composition root constructs them in order:

```typescript
const display = new Display(config, statusBar);
const input = new Input(display);
```

Both are then injected into controllers that need them (`AgentController`, `WorkerController`).

Depends on #706 (agent MVC structure) since both `Display` and `Input` need to be classes with the directory structure in place first.

jasoncrawford:
## Summary

- Adds an `Input` class to `src/agent/views/input.ts` that wraps `Display` as a reference, delegating `ask`, `pick`, `pickMultiple`, and `pickQuestion` to the underlying module-level functions
- Updates `AgentController`, `startWorkerMode`, and `main()` to accept an injected `Input` instance — eliminating the awkward `ask(display.statusBar, ...)` pattern
- Removes `vi.mock` of the input module in tests; instead injects a plain mock object (`mockInput as unknown as Input`) directly into constructors, making mocking explicit and reliable

## Test plan

- [ ] All existing tests pass: `npm test` (1479 tests passing)
- [ ] Type check passes: `npx tsc --noEmit`
- [ ] CI passes on this PR

Closes #715


src/agent/controllers/worker-controller.ts
```diff
@@ -13,7 +13,7 @@ import { Workspace, confirmIfUnsafe } from "../models/workspace.js";
 import { fmtError } from "../../utils.js";
 import { getConfig } from "../../config.js";
 import type { CommandRegistry } from "./command-controller.js";
-import { pick } from "../views/input.js";
+import { pick, Input } from "../views/input.js";
```
jasoncrawford: the `pick` import should not be needed now, right? it should be on `input`

src/agent/views/input.ts
jasoncrawford, line 1: Instead of leaving a bunch of exported standalone functions here, and then creating a class that wraps them, the functions should just move into the class. The intent was to migrate these functions into a class, not to just provide an additional way to access them.

src/agent/index.ts
```diff
@@ -26,11 +26,13 @@ export async function main(
   settings: Settings,
   runWorkerMode?: boolean,
   workspaceCfg?: { workspaceDir: string; repoUrl: string },
+  inp?: Input,
```
jasoncrawford: why is this optional? we should just decide where it is created, either inside or outside main, and stick with it.

jasoncrawford (commented):
All three addressed:

1. **Standalone functions → class only**: Removed `export` from `ask`, `pick`, `pickMultiple`, `pickQuestion`, and `promptLine`. They're now module-private helpers; the `Input` class is the only public API. Added `Input.promptLine()` method so tests can reach it. Updated `repl.autocomplete.test.ts` and `repl.multiline.test.ts` (which also called `ask` directly) to use `testInput.ask()`.

2. **`inp` is now required**: Moved it before the optional `runWorkerMode?` and `workspaceCfg?` params (TypeScript requires required params to precede optional ones). `Input` is created once in the startup block and passed to both `AgentController` and `main()` — no `??` fallback inside `main`.

3. **`pick` import removed from `worker-controller.ts`**: The remaining `pick` reference was the default value for `confirmTaskQuit`'s `pickFn` parameter. Fixed by adding `pickFn?` to `WorkerSessionOptions` — `startWorkerMode` now passes `(opts) => input.pick(opts)` in the options, so production code routes through `Input` and tests continue injecting `mockPick` directly.

src/agent/views/input.ts
jasoncrawford, line 1: It's good that the functions are now private and only the class is exported. But it would be simpler to just define them inline on the class. This structure, where the functions are defined outside the class and then the class just wraps them, is a historical vestige that we can do away with. Write this as if you were writing the Input class for the first time.

src/agent/index.ts
```diff
@@ -24,13 +24,15 @@ export async function main(
   permConfig: AgentPermConfig,
   display: Display,
   settings: Settings,
+  inp: Input,
```
jasoncrawford: Just name this `input` and delete the `const input = inp;` below.

src/agent/index.ts
```diff
@@ -160,10 +162,10 @@ export async function main(
     // empty promptLine suppresses the drawFresh callback so incoming messages
     // are printed cleanly without a prompt preceding or following them.
     const promptStr = session ? (showPrompt ? "\n[agent] > " : "") : "\n> ";
-    const input = await ask(statusBar, promptStr, () => controller.listCommands(), wsAbort);
+    const typed = await input.ask(promptStr, () => controller.listCommands(), wsAbort);
```
jasoncrawford: Maybe call this string `userInput` or `text` or something; `typed` sounds like a reference to static typing

jasoncrawford (commented):
All three addressed:

1. **`ask/pick/pickMultiple/pickQuestion/promptLine` inlined into the class**: Removed all module-level private helper functions. `ask()` now uses `this.display.statusBar` directly. `pickerLine` became a `private static` method. The file reads as if the `Input` class was written from scratch.

2. **Parameter renamed `inp` → `input`**, with the redundant `const input = inp;` line removed.

3. **`typed` → `userInput`** in the routing loop.

PR #727, opened 2026-04-18
refactor: split display.ts into formatters, style, and renderer

### LINKED ISSUE #685: Continue splitting display.ts; resolve event-fmt.ts duplication
display.ts currently has ~807 lines and ~49 exports. The goal is to reduce it to just terminal screen management and printing — abstract/general terminal I/O — with data formatting and content rendering pulled into focused sibling modules.

### 1. `formatters.ts` — pure data-to-string functions

Extract all the data formatting helpers that have no terminal/screen concern:
`fmtCount()`, `fmtTime()`, `fmtTimestamp()`, `fmtDuration()`, `fmtNum()`, `fmtStats()`, `fmtEvent()`, `fmtEventDetails()`, `fmtArgs()`, `toRelativePath()`, `trunc()`, and similar.

These are pure functions (data in, string out) with no dependency on the terminal or the Display class. Putting them in their own module also resolves the long-standing duplication with `src/foreman/event-fmt.ts` — both the agent and the foreman can import `fmtEvent()` / `fmtEventDetails()` from a shared location instead of maintaining two copies. Depending on where the shared module lives (`shared/formatters.ts` or `src/formatters.ts`), `src/foreman/event-fmt.ts` can be deleted or reduced to a thin re-export shim.

### 2. `markdown.ts` (or `renderer.ts`) — rich content rendering

Extract the markdown/diff rendering section:
`renderMarkdown()`, `renderTable()`, `mdInline()`, `fmtHunk()`, `fmtEditResult()`, `fmtBashOutput()`, `fmtWriteOutput()`, `fmtTodoWriteOutput()`, `fmtAskUserQuestionInput()`, `fmtToolSearchOutput()`, `toolResultText()`, `distributeWidths()`, and the `Hunk` type.

This is a self-contained rendering engine. It produces strings from structured data but is independent of terminal screen management.

### 3. `display.ts` — what remains

After both extractions, display.ts should contain only:
- The `Display` class and its print methods
- Terminal layout utilities: `W`, `hr()`, `effectiveWidth()`, `clearBreak()`
- Colors/styles: `c`, `s`
- The format tables (private static maps on `Display`, per #720)
- SDK content block types (`ToolUseBlock`, `ToolResultBlock`, `ContentBlock`)

## Connection to #720

Issue #720 (minimize Display exports) should land alongside or before this — the format tables becoming private static members of `Display` is part of the same cleanup.

jasoncrawford:
## Summary

- **`shared/formatters.ts`** (new) — pure data-to-string helpers: `trunc`, `fmtCount`, `fmtTime`, `fmtDuration`, `fmtNum`, `fmtStats`, `fmtArgs`, `toRelativePath`, `fmtEvent`, `fmtEventDetails`. No terminal/color concerns. Shared by both foreman and agent.
- **`src/agent/views/style.ts`** (new) — terminal color/style constants: `c`, `s`, `W`, `hr`, `effectiveWidth`. Extracted to avoid a circular dependency between `display.ts` and `renderer.ts`.
- **`src/agent/views/renderer.ts`** (new) — rich content rendering engine: `renderMarkdown`, `renderTable`, `fmtHunk`, `fmtEditResult`, `fmtBashOutput`, `fmtWriteOutput`, `fmtTodoWriteOutput`, `fmtTodoWriteInput`, `fmtAskUserQuestionInput`, `fmtToolSearchOutput`, `toolResultText`, `distributeWidths`. Produces strings from structured data, no I/O.
- **`src/agent/views/display.ts`** — slimmed to: `Display` class, `clearBreak`, format tables (private), SDK block types, and `resolve()`. Re-exports `c`, `s`, `W`, `hr`, `effectiveWidth` from `style.ts` for backward compatibility.
- **`src/foreman/models/webhook-event.ts`** — removes ~50 lines of duplicated event-formatting helpers; delegates to `shared/formatters.ts` instead.

Resolves the long-standing duplication between `display.ts` and `webhook-event.ts`: both now import `fmtEvent`/`fmtEventDetails` from a single canonical location in `shared/`.

## Test plan

- [ ] All 1345 unit tests pass (`npm test`)
- [ ] ESLint passes (`npm run lint`)
- [ ] Tests for affected modules: `repl.helpers`, `display.event-details`, `display.relative-paths`, `display.cleanup` updated to import from new locations
- [ ] `display.ts` re-exports ensure no caller changes needed for `c`, `s`, `W`, `hr`, `effectiveWidth`, `clearBreak`

Closes #685


jasoncrawford (commented):
Good structure overall — the three-way split into  (pure data-to-string),  (terminal constants), and  (rich content) is the right direction. A few places where the display/renderer boundary isn't quite clean:

---

### 1. `renderer.ts` calls `getConfig()` directly

```ts
// renderer.ts:66
export function fmtBashOutput(text: string): string {
  const t = text.trim();
  if (!t || t === "(Bash completed with no output)") return "Success";
  return getConfig().verbose ? t : trunc(t, 100);  // ← renderer reading runtime config
}
```

The renderer's job is to produce strings from data — it shouldn't need to know about display mode. The `verbose` decision is a display concern, not a rendering concern. Fix: accept `verbose` as a parameter, or always return the full text and let the caller truncate if it wants:

```ts
export function fmtBashOutput(text: string, verbose = false): string {
  const t = text.trim();
  if (!t || t === "(Bash completed with no output)") return "Success";
  return verbose ? t : trunc(t, 100);
}
```

---

### 2. Re-exports from `display.ts` for backward compatibility

```ts
// display.ts:34 — after the split, these now live in style.ts
export { c, s, W, hr, effectiveWidth };
```

The comment says "so importers that previously sourced them from display.ts continue to work" — but that's the re-export anti-pattern. When splitting a file, the right move is to update callers directly, not to paper over the split with re-exports. Re-exports make the split look complete while leaving old coupling intact. Update every caller to import from `style.ts` instead, then remove this line.

---

### 3. `style.ts` calls `getConfig()` inside `effectiveWidth()`

```ts
// style.ts:21
export function effectiveWidth(fallback = W): number {
  return (process.stdout.columns ?? fallback) - (getConfig().verbose ? VERBOSE_PREFIX_LEN : 0);
}
```

The verbose timestamp prefix is a display rendering decision — the fact that verbose mode adds a 9-char prefix to every line is something Display owns. Style constants shouldn't be calling into runtime config. Options: (a) move `effectiveWidth` into `display.ts` (or remove the standalone export entirely, since `Display` already has `this.effectiveWidth()`), or (b) pass `verbose` as a parameter so `style.ts` stays dependency-free.

---

### 4. Format tables in `display.ts` bypass the verbose/quiet dispatch mechanism with inline `getConfig()` calls

```ts
// display.ts:90
thinking: (b) => c.gray("\n" + (getConfig().thinkOutLoud ? renderMarkdown(b.thinking ?? "") : "Thinking...")),
// display.ts:116
_default: (b) => c.darkGray(`→ ${getConfig().verbose ? toolResultText(b) : trunc(toolResultText(b), 100)}`),
```

The format table already has a verbose/quiet dispatch mechanism — entries can be `{ quiet: Fmt, verbose: Fmt }`. These two entries are doing their own verbose check inline instead. Use the mechanism that's already there:

```ts
thinking: {
  quiet:   (b) => c.gray("\nThinking..."),
  verbose: (b) => c.gray("\n" + renderMarkdown(b.thinking ?? "")),
},
// and for thinkOutLoud specifically, Display.resolve() dispatches on config.verbose —
// if thinkOutLoud should differ from verbose, it needs its own mechanism,
// but the current code ignores the quiet/verbose dispatch entirely

_default: {
  quiet:   (b) => c.darkGray(`→ ${trunc(toolResultText(b), 100)}`),
  verbose: (b) => c.darkGray(`→ ${toolResultText(b)}`),
},
```

Note: `thinkOutLoud` and `verbose` are separate config fields. If `thinkOutLoud` should behave differently from `verbose` for the verbose/quiet table dispatch, that's worth a separate conversation — but either way the inline `getConfig()` call should go.

src/agent/views/display.ts
```diff
@@ -682,7 +243,7 @@ export class Display {
 
   /** Returns the usable terminal width, adjusted for verbose timestamp prefix. */
   effectiveWidth(fallback = W): number {
-    return (process.stdout.columns ?? fallback) - (this.config.verbose ? VERBOSE_PREFIX_LEN : 0);
+    return (process.stdout.columns ?? fallback) - (this.config.verbose ? 9 : 0);
```
jasoncrawford: What? this was a named constant, now it's a magic number. That's not good. This should not have been changed

jasoncrawford: Also why is this defined here and in style.ts? Seems like it should only be in one place, and probably in renderer.ts

src/agent/views/display.ts
```diff
@@ -2,59 +2,43 @@ import * as Wire from "../../../shared/wire.js";
...
+
+// ── clearBreak ─────────────────────────────────────────────────────────────
 
 export function clearBreak(): string {
-  const width = effectiveWidth();
+  const width = effectiveWidth(W, getConfig().verbose);
   const label = "=== Context cleared ";
   const fill = "=".repeat(Math.max(0, width - label.length));
   return "\n" + c.sageGreen(s.bold(label + fill));
 }
```
jasoncrawford: seems like clearBreak could go in renderer.ts?

src/agent/views/display.ts
```diff
@@ -86,445 +62,28 @@ export type ContentBlock =
...
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type Fmt = (data: any) => string | null;
+export type FmtEntry = Fmt | { quiet?: Fmt; verbose?: Fmt };
+export type FmtTable = Record<string, FmtEntry>;
 
 function fmtToolCall(b: ToolUseBlock, fmt: string) {
   fmt = c.skyBlue(`\n${fmt}`);
   if (b.input?.description) fmt += c.gray(` # ${b.input.description}`);
   return fmt;
 }
```
jasoncrawford: this could go in renderer?

src/agent/views/display.ts
```diff
@@ -86,445 +62,28 @@ export type ContentBlock =
...
-    out.push(mdInline(line));
-  }
-
-  if (inCode && codeLines.length) out.push(codeLines.map(l => "  " + l).join("\n"));
-  flushTable();
-  return out.join("\n");
+function fmtSubagentType(subagentType: string | null | undefined): string {
+  if (!subagentType || subagentType === "general-purpose") return "Subagent";
+  return subagentType;
 }
```
jasoncrawford: renderer?

src/agent/views/display.ts
```diff
@@ -86,445 +62,28 @@ export type ContentBlock =
...
-// ── FORMATS ───────────────────────────────────────────────────────────────────
-
-// FmtTable is a mixed-type dispatch table: each entry receives a different
-// runtime-typed value, so `any` is intentional here rather than `unknown`.
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export type Fmt = (data: any) => string | null;
-export type FmtEntry = Fmt | { quiet?: Fmt; verbose?: Fmt };
-export type FmtTable = Record<string, FmtEntry>;
-
 const ASSISTANT_BLOCK_FMT: FmtTable = {
```
jasoncrawford: all of these FmtTables, and the things that invoke them, could go in renderer?

jasoncrawford (commented):
The intended split between Display and Renderer is:

> Renderer = produces strings. Doesn't know about the terminal, doesn't write to stdout, doesn't know about the status bar. renderMarkdown(text) → string. Pure, testable in isolation.
> 
> Display = manages terminal I/O. Knows about the status bar (clears it before printing, redraws it after), handles verbose-mode line prefixing, is the single doorway through which everything reaches stdout.

Let's implement that more cleanly.

src/agent/views/display.ts
```diff
@@ -1,674 +1,48 @@
...
+  type FmtTable,
+  clearBreak as _clearBreak,
+  ASSISTANT_BLOCK_FMT,
+  USER_BLOCK_FMT,
+  TOOL_CALL_FMT,
+  TOOL_RESULT_FMT,
+  TOOL_ERROR_FMT,
+  SYSTEM_FMT,
+  MESSAGE_FMT,
+  FOREMAN_MESSAGE_FMT,
```
jasoncrawford: all these _FMT maps should be private and internal to renderer.

src/agent/views/style.ts
```diff
@@ -0,0 +1,49 @@
...
+export const VERBOSE_PREFIX_LEN = 9;
+
+/**
+ * Returns the usable terminal width, accounting for the verbose timestamp
+ * prefix when verbose mode is active. Pass `verbose` from the caller's
+ * config rather than reading it here so this module stays dependency-free.
+ */
+export function effectiveWidth(fallback = W, verbose = false): number {
+  return (process.stdout.columns ?? fallback) - (verbose ? VERBOSE_PREFIX_LEN : 0);
+}
```
jasoncrawford: As far as I can tell, this is only used in renderer, why not make it private there?

jasoncrawford (commented):
See feedback below. Note that there are merge conflicts also

PR #728, opened 2026-04-18
refactor: create WorkspaceController class; move workspace lifecycle out of WorkerController

### LINKED ISSUE #725: Create WorkspaceController class; move workspace concerns out of WorkerController
Two problems with the current workspace controller setup:

1. `workspace-controller.ts` exports only a `registerWorkspaceCommands()` function — no `WorkspaceController` class. It's a controller in name only.
2. `worker-controller.ts` contains workspace lifecycle concerns (creating the workspace, resetting between tasks, the `afterTask` hook) that have nothing to do with the WebSocket/foreman protocol.

## Fix

Create a proper `WorkspaceController` class in `workspace-controller.ts`:

```typescript
export class WorkspaceController {
  constructor(private workspace: Workspace, private display: Display) {}
  registerCommands(registry: CommandRegistry): void { ... }
  async onCreate(): Promise<void> { ... }
  async onReset(): Promise<void> { ... }
}
```

Move workspace lifecycle logic from `worker-controller.ts` into `WorkspaceController`. `WorkerController` should only handle the WebSocket/foreman protocol — connecting, receiving messages, managing the task loop.

## Wiring

`index.ts` (the composition root) constructs both `Workspace` (the model) and `WorkspaceController`, injects `Workspace` and `Display` into the controller, and calls `workspaceController.registerCommands(registry)` alongside the other controllers. `WorkerController` receives a `WorkspaceController` reference (or just the specific callbacks it needs) rather than owning workspace operations directly.

Depends on #706 (agent MVC structure) being in place, which it is.

jasoncrawford:
## Summary

- Creates a proper `WorkspaceController` class in `workspace-controller.ts` with `registerCommands()`, `onCreate()`, and `onReset()` methods
- `onCreate()` owns workspace event subscriptions + `create()` + `process.chdir()` (previously scattered in `startWorkerMode`)
- `onReset()` is the `afterTask` callback — `confirmIfUnsafe` + `workspace.reset()`, throwing on cancel/failure (previously an inline closure in `startWorkerMode`)
- `startWorkerMode()` in `worker-controller.ts` now delegates to `WorkspaceController`, keeping it focused on the WebSocket/foreman protocol
- All callers (`index.ts`, tests) updated to use `new WorkspaceController(workspace, display).registerCommands(registry)` instead of the old standalone `registerWorkspaceCommands` function
- CLAUDE.md updated to reflect the new class

## Test plan

- [ ] All existing tests pass (`npm test` shows 1345/1345 passing)
- [ ] Type check clean (`npx tsc --noEmit`)
- [ ] Workspace slash command tests in `repl.workspace.test.ts` and `worker.test.ts` use `WorkspaceController` directly

Closes #725


src/agent/controllers/worker-controller.ts
jasoncrawford, line 1: I still see some things in this file where the WorkerController is managing the workspace. Review it again and see if there is anything that can be cleanly delegated to the WorkspaceController.

src/agent/index.ts
```diff
@@ -82,7 +82,7 @@ export async function main(
   // modes; commands that require a foreman connection degrade gracefully.
   const registry = new CommandRegistry();
   const controller = new CommandController(registry);
-  registerWorkspaceCommands(workspace, registry.scoped("workspace"), display);
+  new WorkspaceController(workspace, display).registerCommands(registry.scoped("workspace"));
```
jasoncrawford: This creates a WorkspaceController, uses it to register commands, then throws it away. Later, the WorkerController creates its own, different WorkspaceController.

I think all the controllers should be instantiated exactly once in index.ts, and then injected into each other as needed.

jasoncrawford (commented):
Good catches — both issues addressed in the follow-up commit.

**Single WorkspaceController instance**: `WorkspaceController` is now constructed once in `index.ts` (from `workspaceCfg`) and passed to `startWorkerMode`. The session receives `workspaceController` in its options (replacing the old `workspace` field). The slash command tests in `worker.test.ts` now construct `WorkspaceController` directly rather than going through the session.

**Remaining workspace concerns in worker-controller.ts**: `startWorkerMode` previously had three separate workspace destroy/confirm calls (in `shutdown`, SIGTERM handler, and `cleanup`). These are now delegated to two new methods on `WorkspaceController`:
- `onDestroy()` — confirms if unsafe, then destroys (used in `shutdown` and `cleanup`)  
- `onForceDestroy()` — destroys immediately without prompting (used on SIGTERM)

`worker-controller.ts` no longer imports `Workspace`, `confirmIfUnsafe`, `path`, or `os` — all workspace-specific logic lives in `WorkspaceController`.

PR #740, opened 2026-04-19
refactor: replace sentinel string signals with typed WsInputSignal

### LINKED ISSUE #733: Replace sentinel string signals with typed worker signals
## Problem

`WorkerSession` signals `index.ts` by resolving promises with hidden sentinel strings (`__ws_prompt__`, `__ws_fatal__`), then exposing `isWsSignal()`/`isFatalSignal()` statics to classify them. The routing loop does string-matching to detect what are really typed events.

## Fix

Change `createWsInputPromise()` to return a typed discriminated union (e.g. `{ type: 'ws_prompt' | 'ws_fatal' | 'user_input'; value?: string }`). The `is*` static methods go away.

jasoncrawford:
## Summary

- Replaces `__ws_prompt__` / `__ws_fatal__` hidden sentinel strings with a typed discriminated union `WsInputSignal = { type: "ws_prompt" } | { type: "ws_fatal" }`
- `createWsInputPromise()` now returns `Promise<WsInputSignal>` instead of `Promise<string>`
- Removes `WorkerSession.isWsSignal()` and `WorkerSession.isFatalSignal()` static methods; callers in `index.ts` discriminate on `.type` directly
- Makes `Input.ask()` generic (`ask<TAbort = never>`) so the abort promise value flows through typed

## Test plan

- [x] All existing tests pass (`npm test` — 1341 tests, 0 failures in non-DB tests)
- [x] Updated `worker.test.ts` tests to assert on `.type` property instead of removed static methods

Closes #733


jasoncrawford:
Closing without merging — this fixes the symptom (untyped sentinels) but not the root cause. The real fix is to decouple WorkerSession from Input.ask() so sentinels aren't needed at all. See #744.

Issue #418, opened 2026-03-30
Something is randomly inserting blank lines


jasoncrawford:
Sometimes when nothing is happening with a worker, or in the middle of an agent loop, blank lines will randomly get printed on the console. No idea where this is coming from. Doesn't fit any obvious pattern. Something is just deciding to print a newline.

PR #748, opened 2026-04-19
refactor: decouple Input from WorkerSession — event-driven routing loop

### LINKED ISSUE #744: refactor: decouple Input from WorkerSession — event-driven routing loop
## Problem

The routing loop in `index.ts` has an awkward design rooted in the fact that `Input.ask()` serves two purposes: receiving user input *and* acting as the rendezvous point for foreman signals.

```ts
// current: ask() takes an abort promise that resolves to sentinels
const wsAbort = session?.createWsInputPromise();
const userInput = await input.ask(promptStr, () => controller.listCommands(), wsAbort);

if (userInput === "__abort__") continue;
if (WorkerSession.isFatalSignal(userInput)) { ... }
if (WorkerSession.isWsSignal(userInput)) { ... }
// only now do we know userInput is an actual user string
```

Specific smells:
- `ask()` takes `abort?: Promise<string>` — a promise that resolves to arbitrary sentinel strings mixed into the return type
- `WorkerSession.createWsInputPromise()` is a weird, stateful one-shot promise that exists solely to thread a signal into `ask()`
- Sentinel strings (`__ws_prompt__`, `__ws_fatal__`, `__abort__`, `__eof__`) are all crammed into the same `string` return type
- `showPrompt` is a mutable boolean in the loop that controls whether to pass an empty or non-empty prompt string to `ask()` — a hack to suppress the TUI prompt when waiting for a task
- The logic for clearing the prompt when a foreman message arrives lives *inside* `ask()` (in the abort handler), not where it belongs (the controller layer)

## Design

The right model is: there are two independent input sources — the user's keyboard and the foreman WebSocket. A router (in `index.ts` or a new controller class) listens to both and dispatches to the right handler. Neither source should know about the other.

### Changes

**`Input`**
- Remove the `abort?: Promise<string>` parameter from `ask()`.
- Add `Input.cancel(): void` — clears the prompt area and resolves the current `ask()` with `null` (or a typed "cancelled" indicator), so the router can interrupt user input when a foreman event arrives.

**`WorkerSession`**
- Extend `EventEmitter`. Emit a typed event (e.g. `"prompts_ready"`) when a prompt is queued, and `"fatal"` on a fatal error, instead of resolving a promise with a sentinel string.
- Remove `createWsInputPromise()`, `isWsSignal()`, `isFatalSignal()`.

**Routing loop (`index.ts`)**
- Subscribe to `session` events once at startup (not every loop iteration).
- When `"prompts_ready"` fires: call `input.cancel()` to clear any live user prompt, run the queued prompts, then restart the user prompt.
- When `"fatal"` fires: cancel user input, drop back to interactive mode.
- Replace the `showPrompt` boolean with the simpler invariant: if we're actively waiting for user input, `ask()` is running; if not, it isn't. No empty-string prompt hack.
- `ask()` return type is `string | null` — `null` means "was cancelled by the router" and the loop just continues.

The result is a loop whose shape is: *await user input; when foreman interrupts, handle it and loop back*. The `ask()` function becomes a pure input primitive with no knowledge of the foreman.

jasoncrawford:
## Summary

- `WorkerSession` now extends `EventEmitter`, emitting `"prompts_ready"` when a prompt is enqueued and `"fatal"` on a fatal foreman error — replacing the old sentinel-string mechanism (`createWsInputPromise` / `isWsSignal` / `isFatalSignal`)
- `Input.ask()` removes the `abort?: Promise<string>` parameter and changes its return type to `Promise<string | null>`. The new `Input.cancel()` method resolves any active `ask()` with `null` and clears the prompt area
- The routing loop in `index.ts` subscribes to session events once at startup (both call `input.cancel()`), replaces the `showPrompt` boolean with always-visible prompts, and handles `null` returns from `ask()` by draining queued prompts

## Test plan

- [x] All existing tests pass (1341 tests, only the 3 Supabase-requiring DB tests fail as before)
- [x] New event-based tests cover `"prompts_ready"` and `"fatal"` emission
- [x] New `cancel()` tests in `repl.input.test.ts` verify null return, no-op before/after active ask, and the terminal escape sequence invariants from issue #418 (no `\r\n\x1b[J`, cursor-up to clear `\n` prefix)
- [x] `npx tsc --noEmit` and `npm run lint` pass clean

Closes #744


jasoncrawford (commented):
This is much cleaner, big step in the right direction. However, input.ts still has this core loop where it awaits on `ask()`, and then it has to interpret a null result as maybe meaning pending prompts. Is there a way for the core routing code to just listen to *both* input and any active worker session, and then to go to sleep and listen for events? I'm thinking there should be a more generalized event-driven architecture here, rather than this historical REPL architecture.

jasoncrawford (commented):
Good catch — the `null` return was still a subtle side-channel signal. I've pushed a follow-up commit that replaces the global `session.on()` subscriptions with `Promise.race` in the routing loop itself.

Each iteration now explicitly races two sources:

```typescript
const winner = await Promise.race([
  askPromise.then((v) => ({ from: "input", value: v })),
  sessionPromise.then((ev) => ({ from: "session", event: ev })),
]);

if (winner.from === "session") {
  input.cancel();       // release the terminal
  await askPromise;     // wait for cleanup (resolves null)
  await drainPendingPrompts();
  continue;
}
```

The routing loop now reads as "wait for either user input or a session event" rather than "wait for input, and hope a session event secretly cancels it and returns null." The `cancel()` call is still needed to clean up the terminal display when the session wins, but it's now an explicit action rather than an implicit side-effect of a global subscription.

The one remaining constraint preventing `Input` from being a pure event emitter is that `ask()` owns the terminal — it renders the prompt, handles autocomplete, and manages cursor state. There would need to be a separation between "start accepting input" and "await the next line" before `Input` could be a proper event source that can be raced without this explicit cancel/cleanup step. That feels like a natural future issue if we want to go further.

jasoncrawford (commented):
Using a promise race is an interesting idea, and reasonably clean. However, it creates a lot of machinery in awaitSessionEvent to turn an event emitter into a promise. If there are more new types of events, we'll have to add more handlers, more setup, more cleanup, etc.

What if instead we inverted this, and made it event-driven instead of promise-driven. No machinery is needed for the worker session. Instead, we just need a core loop that listens for input from `ask()`, and fires an event when there is real input. (If `ask` has to return null/undefined sometimes when it is interrupted, that's fine, we just don't fire a user input event in that case).

Then the core command/request routing infrastructure is event-driven.

jasoncrawford (commented):
Good call. I've pushed a new approach: rather than converting session events to promises (so they can be raced), I inverted it — both sources push to a shared async FIFO channel, and the routing loop just `await`s the next item:

```typescript
// Session events push directly to the channel (and cancel any active prompt).
session.on("prompts_ready", () => { input.cancel(); enqueueRoutingEvent(...); });
session.on("fatal",         () => { input.cancel(); enqueueRoutingEvent(...); });

// ask() pushes when the user submits real input; null (cancelled) is ignored.
const listenForInput = () => {
  void input.ask(promptStr, getCommands).then((line) => {
    if (line !== null) enqueueRoutingEvent({ type: "line", value: line });
  });
};

while (true) {
  if (session?.hasPendingPrompts()) { await drainPendingPrompts(); continue; }
  listenForInput();
  const event = await nextRoutingEvent(); // sleeps until something arrives
  if (event.type === "session") { await drainPendingPrompts(); continue; }
  // handle event.value as user input...
}
```

No `awaitSessionEvent` machinery. Adding new event types is just adding another `session.on(...)` that calls `enqueueRoutingEvent` — no setup/cleanup boilerplate. The `input.cancel()` call is still there (terminal cleanup is unavoidable when a session event interrupts user input), but it's a plain one-liner on each handler rather than infrastructure.

PR #751, opened 2026-04-19
refactor: clean up index.ts composition root / main() split

### LINKED ISSUE #750: refactor: clean up index.ts composition root / main() split
## Problem

`index.ts` has an unprincipled split between the startup block and `main()`. Currently:

- **Startup block** constructs: `settings → statusBar → display → input → picker → agentController`, then wraps `agentController.runQuery` in a closure to produce `runQueryFn`, then calls `main()`
- **`main()`** receives those objects, then constructs more: `workspace → workspaceController → workerCtx/session → settingsController → registry → controller`

There's no clear rule governing what belongs where. The `runQueryFn` closure is a smell — `RunQuery` is essentially a single-method interface, and the closure just adapts `AgentController` to it.

## Solution

Make the startup block the single composition root: it builds the full object graph, `main()` just runs the routing loop.

Concretely:
- Pass `agentController` (or a `QueryRunner` interface) directly to `main()` instead of wrapping it in a `runQueryFn` closure
- Move the remaining construction from inside `main()` into the startup block, or otherwise establish a clear principle: build everything first, then run
- Drop the `RunQuery` typedef and the wrapping closure

**Principle:** one place constructs, one place runs. `main()` should accept a fully-built set of deps and loop — not construct more objects mid-flight.

jasoncrawford:
## Summary

- Drop the `RunQuery` typedef from `worker-controller.ts` and the wrapping closure from the startup block — `agentController` is now passed directly to `main()`
- Move all object construction (`Workspace`, `WorkspaceController`, `startWorkerMode`, `fetchModelsFn`, `SettingsController`, `CommandRegistry`, `CommandController`) from inside `main()` into the startup block — the startup block is now the single composition root
- `main()` accepts a fully-built object graph, registers commands (command handlers still close over loop-local `sessionId`/`doExit`), and runs the routing loop — no `new X()` calls mid-flight
- Update `worker.main.test.ts` with a `buildWorkerDeps()` helper that mirrors the startup block's construction, replacing the old `runWorkerMode`/`workspaceCfg` convenience params
- Update `CLAUDE.md` to document the new principle

## Test plan

- [ ] `npm test` — all 1343 tests pass

Closes #750


jasoncrawford (commented):
This is a cleaner split. But now main takes 12 arguments.

Suggestion: Declare a BrunelAgent class, with instance methods. Put all the setup in a constructor, except for the stuff that depends on runWorkerMode. Move anything that depends on runWorkerMode into main. Rename main() to start(runWorkerMode), it takes runWorkerMode as a parameter.

Then the only code outside of the class parses the --worker-mode parameter, creates a new BrunelAgent(), and invokes start(runWorkerMode).

jasoncrawford (commented):
Done — pushed a follow-up commit that introduces `BrunelAgent` as you suggested.

**Constructor** builds the full synchronous object graph (settings → statusBar → display → input → picker → agentController → confirm → workspace → workspaceController → fetchModelsFn → settingsController → registry → controller).

**`start(runWorkerMode)`** handles the async part (calls `startWorkerMode` if needed), registers commands (which still need to happen here since some handlers close over start()-local `sessionId` and `doExit`), then runs the routing loop.

**Startup block** is now just three lines: load config → parse `--worker-mode` → `new BrunelAgent(config).start(runWorkerMode)`.

For tests, module-level `vi.mock` for `Input`, `Picker`, and `AgentController` lets the constructor pick up test doubles — a small `installMocks(runQueryFn)` helper wires them before each `new BrunelAgent(config)` call.

PR #758, opened 2026-04-19
fix: forward issues/closed and issues/reopened events to worker; restructure routeIssueEvent()

### LINKED ISSUE #755: fix+refactor: forward issues/closed and issues/reopened events to worker; restructure routeIssueEvent()
## Bug

`routeIssueEvent()` in `wss.ts` never forwards `issues/closed` or `issues/reopened` events to the worker. Both branches call the task manager then **return early**, skipping `forwardEvent` at the bottom of the method.

This means the worker's `issueClosed` tracking code in `handleMessage()` is dead — the flag is never set. As a result, when a user quits the worker after the GitHub issue has been closed, `confirmTaskQuit()` always shows "task is still open, quit anyway?" instead of "task is closed, complete before exiting?"

## Root cause

The method mixes two concerns — foreman-side effects and worker notification — with some actions returning early and some falling through to `forwardEvent`. The rule governing which actions forward is implicit. `closed` and `reopened` were structured like `unlabeled` (which correctly doesn't forward), but shouldn't be.

## Fix

Restructure into two explicit phases:

**Phase 1 — foreman-side effects.** Return early only when there's genuinely nothing to forward: no task found, or `unlabeled` (worker doesn't need to know the label was removed).

**Phase 2 — worker notification.** A single `forwardEvent` at the bottom, reached by everything that didn't return early.

```ts
// Phase 1: foreman-side effects
if (!task) { /* maybe enqueue */ return routeResult(...); }
if (action === "unlabeled") { dequeue; assignWork; return routeResult(task); } // explicit: no forward
if (action === "closed")    { closeIssue; assignWork; }
if (action === "reopened")  { reopenIssue; assignWork; }
if (action === "edited")    { handleBodyEdited; }

// Phase 2: forward to worker
forwardEvent(task, evt, `#${issueNumber}`);
return routeResult(task);
```

**Forwarding is now the default; skipping it requires an explicit `return`.** The bug is structurally impossible to reintroduce accidentally.

jasoncrawford:
## Summary

- `issues/closed` and `issues/reopened` events now fall through to `forwardEvent` instead of returning early after foreman-side effects
- `routeIssueEvent` is restructured into two explicit phases: foreman effects first, then a single `forwardEvent` at the bottom
- Explicit `return` statements are now reserved for cases with nothing to forward (`unlabeled`, enqueue path), making it structurally impossible to accidentally skip forwarding again

## Root cause

Both `closed` and `reopened` branches called `return routeResult(task)` after calling the task manager, skipping the `forwardEvent` call at the bottom. The worker's `issueClosed` flag was never set, so `confirmTaskQuit()` always showed "task is still open, quit anyway?" even when the issue was closed.

## Test plan

- [x] Added `issues/closed is forwarded to the assigned worker` test
- [x] Added `issues/reopened is forwarded to the assigned worker` test
- [x] All 1345 unit tests pass

Closes #755

### DIFF
--- a/src/foreman/controllers/wss.ts
+++ b/src/foreman/controllers/wss.ts
@@ -526,8 +526,12 @@ export class ForemanWss {
 
     const action = p.action as string | undefined;
 
-    // If the issue isn't queued yet, check if this webhook should enqueue it.
+    // Phase 1: foreman-side effects.
+    // Return early only when there is genuinely nothing to forward: no task tracked,
+    // or "unlabeled" (worker doesn't need to know the label was removed).
+
     if (!task) {
+      // Check if this webhook should enqueue a new task.
       const labeledNow =
         action === "labeled" &&
         (p.label as R | undefined)?.name === this.config.taskLabel;
@@ -551,12 +555,15 @@ export class ForemanWss {
         log(`[task #${issueNumber}] enqueued via issues/${action}`);
         return routeResult(enqueued);
       }
+      // No task found and not an enqueue event — still fall through so closed/reopened
+      // can update blocker state (issue may be a dependency, not a task itself).
     }
 
     if (
       action === "unlabeled" &&
       (p.label as R | undefined)?.name === this.config.taskLabel
     ) {
+      // Explicit early return: worker doesn't need to know the label was removed.
       try {
         await this.taskManager.dequeueIssue(issueNumber);
         log(`[task #${issueNumber}] dequeued (label removed)`);
@@ -576,7 +583,6 @@ export class ForemanWss {
         return routeResult(task);
       }
       await this.assignWork();
-      return routeResult(task);
     }
 
     if (action === "reopened") {
@@ -587,7 +593,6 @@ export class ForemanWss {
         return routeResult(task);
       }
       await this.assignWork();
-      return routeResult(task);
     }
 
     if (action === "edited") {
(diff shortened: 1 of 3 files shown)


src/foreman/controllers/wss.ts
jasoncrawford, line 1: All the catch blocks early-exit, is that intentional/correct?

jasoncrawford (commented):
A way to make this even clearer would be to split phase 1 and phase 2 into two separate methods. That way phase 2 always gets invoked. If you want to skip it, you have to return something explicit from phase 1, like a null or something.

jasoncrawford (commented):
Done — split into `applyIssueEffects` (phase 1) and `routeIssueEvent` (phase 2) in the latest commit. Phase 2 is now always invoked; returning `null` from phase 1 is the only way to suppress forwarding.

On the catch blocks: yes, intentional. If the foreman fails to record a state change (e.g. `closeIssue` throws), notifying the worker of that change would create inconsistency — the worker would think the issue is closed while the foreman's state disagrees. So all catch paths return `null` to skip forwarding. I've added a comment in the method doc to make this explicit.

src/foreman/controllers/wss.ts
```diff
@@ -600,8 +607,13 @@ export class ForemanWss {
       }
     }
 
-    if (!task) return routeResult(null);
-    this.forwardEvent(task, evt, `#${issueNumber}`);
+    return task; // null if no task (e.g., dependency issue — nothing to forward)
+  }
+
+  /** Phase 2: always invoked; forwards to the worker iff phase 1 found a task to notify. */
```
jasoncrawford: This is good but now this comment isn't quite right, this method isn't phase 2, it's the whole thing. Phase 2 is now short enough it doesn't need its own explanation.

PR #767, opened 2026-04-19
test: encapsulate terminal width — remove process.stdout.columns patching

### LINKED ISSUE #765: test: encapsulate terminal width in tests — remove fragile process.stdout.columns patching
## Problem

Several display tests mock terminal width by patching the global `process.stdout.columns` via `Object.defineProperty`. This is fragile: it mutates global state, can leak between tests if cleanup fails, and is brittle across environments.

Example pattern currently used:
```ts
const originalColumns = process.stdout.columns;
Object.defineProperty(process.stdout, "columns", { value: 80, configurable: true });
// ... test ...
Object.defineProperty(process.stdout, "columns", { value: originalColumns, configurable: true });
```

## Fix

`Display.effectiveWidth()` already encapsulates the width calculation. Tests that need a specific width should construct `Display` with a mock config that has a predictable width, or pass width directly to the functions under test. If `effectiveWidth()` were injectable or overridable in tests, the global patching would be unnecessary.

Options:
- Make `Display` accept an optional `getColumns` function (defaults to `() => process.stdout.columns`) injected in the constructor — tests pass `() => 80`
- Or accept a `columns` override in the test config

Either way, eliminate `Object.defineProperty` on `process.stdout` from the test suite.

jasoncrawford:
## Summary

- Add `getColumns: () => number | undefined` as an injectable property on `StatusBar` and `Display` (defaults to `() => process.stdout.columns`)
- Update `Input._screenPosOf()` and `Input._renderSuggestions()` to call `this.display.getColumns()` instead of reading `process.stdout.columns` directly
- Replace all `Object.defineProperty(process.stdout, 'columns', ...)` calls in tests with simple property assignments (`bar.getColumns = () => 120`, `testDisplay.getColumns = () => 80`, etc.)

## Test plan

- [ ] `npm test` — all 1347 existing tests pass, no new failures
- [ ] Grep for `Object.defineProperty.*stdout.*columns` in tests — zero matches

Closes #765

### DIFF
--- a/src/agent/views/status-bar.ts
+++ b/src/agent/views/status-bar.ts
@@ -70,9 +70,16 @@ export class StatusBar extends EventEmitter {
   // prompt area including any leading blank line (see issue #418).
   inputClear: (() => void) | null = null;
 
-  constructor({ agentId, settings, ...initial }: { agentId: string; settings?: Settings } & Omit<WorkerStatusPatch, "reconnectAt">) {
+  /**
+   * Returns the current terminal column count. Defaults to reading
+   * `process.stdout.columns`; override in tests to avoid global patching.
+   */
+  getColumns: () => number | undefined = () => process.stdout.columns;
+
+  constructor({ agentId, settings, getColumns, ...initial }: { agentId: string; settings?: Settings; getColumns?: () => number | undefined } & Omit<WorkerStatusPatch, "reconnectAt">) {
     super();
     this.agentId = agentId;
+    if (getColumns) this.getColumns = getColumns;
     if ("connectionStatus" in initial) this._connectionStatus = initial.connectionStatus!;
     if ("disconnectCode" in initial) this._disconnectCode = initial.disconnectCode;
     if ("taskNumber" in initial) this._taskNumber = initial.taskNumber;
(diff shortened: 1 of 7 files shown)


src/agent/views/status-bar.ts
```diff
@@ -70,9 +70,16 @@ export class StatusBar extends EventEmitter {
   // prompt area including any leading blank line (see issue #418).
   inputClear: (() => void) | null = null;
 
-  constructor({ agentId, settings, ...initial }: { agentId: string; settings?: Settings } & Omit<WorkerStatusPatch, "reconnectAt">) {
+  /**
+   * Returns the current terminal column count. Defaults to reading
+   * `process.stdout.columns`; override in tests to avoid global patching.
+   */
+  getColumns: () => number | undefined = () => process.stdout.columns;
```
jasoncrawford, line 77: does StatusBar have a reference to Display, like Renderer? do we have to reimplement this?

jasoncrawford: No — the dependency is inverted here: `Display` takes `StatusBar` in its constructor, so `StatusBar` has no reference to `Display`. Adding one would create a circular dependency. That's why `StatusBar` needs its own `getColumns` property rather than delegating to `Display`.

PR #769, opened 2026-04-19
fix: always cancel active ask() when processing a queued session event

### LINKED ISSUE #761: User input got sent to agent multiple times
Strange sequence of events just happened.

Worker finished a PR, it was merged. Here was the followup:

```
Summary:

PR #758 is merged. Local branch deleted. Here's a recap of what was done and one thing worth knowing:

What was delivered:
• Bug fix: issues/closed and issues/reopened events are now forwarded to the assigned worker, enabling the worker's issueClosed flag to be set correctly so confirmTaskQuit() shows the right prompt.
• Refactor: routeIssueEvent was split into applyIssueEffects (private — foreman side effects) and routeIssueEvent (public — always invokes forwarding). Returning null from the private method is the only way to suppress notification, making accidental omission structurally impossible.

One minor known regression worth knowing:
The issues/labeled webhook event's task_id is no longer stored in the webhook_events DB table (it was previously linked). The admin dashboard won't show which task was created by a given labeled event. This was a tradeoff to avoid accidentally queuing the labeled event to the worker on connect (which broke the pipeline test). If this association matters for tracing, it could be restored by logging the task separately — worth filing as a follow-up issue if desired.

No new skills or CLAUDE.md updates are needed from this work.

40s, 6 turns, tokens: 8 in / 1.36k out

[agent] > I see. It's probably better to have that association. Is there an actual problem with forwarding it to the worker? The worker will ignore it, right? 
```

That last line was my feedback. After that, the worker immediately started working on code changes in response. Somehow a lot of blank lines got printed after:

```
[agent] > I see. It's probably better to have that association. Is there an actual problem with forwarding it to the worker? The worker will ignore it, right? 







The user is asking whether forwarding the issues/labeled event to the worker would actually cause a problem. They're suggesting we should restore the task ID association in the DB log by returning the enqueued task from applyIssueEffects
```

That last line was the beginning of an agent loop. It ended in a new PR. After that, though, I see a series of agent messages that seem to indicate that the previous user input (starting with “I see.…”) was sent repeatedly to the agent:

```
PR open at https://github.com/jasoncrawford/brunel/pull/760.

You were right — the worker ignores labeled events entirely (only closed and reopened have handlers). So there's no problem forwarding it. The fix restores the task_id association in the DB log and updates the pipeline test to consume the extra event_notification that arrives flushed alongside task_assigned.

3m5s, 16 turns, tokens: 18 in / 9.03k out

[agent] > our-5e3880f3 ∙ sonnet ∙ task #755 ∙ PR #758 ∙ fix-enqueue-task-id-association                                                                                                                                            Connected
The user is asking about the same question again - they may not have seen my previous response. Let me check if PR #760 was already created and summarize what I did.

Actually, looking at the conversation - the user is asking this question, but I already created PR #760 in my previous response. Let me just confirm what I already did.

Yes, exactly right — and I've already fixed it. PR #760 is open with the restoration. The worker's issues handler only acts on closed and reopened; labeled falls through with no effect.

The fix in that PR:
• applyIssueEffects returns the enqueued task again (restoring task_id in the DB log)
• pipeline.test.ts updated to consume the flushed issues/labeled event notification that arrives right after task_assigned

9s, 1 turn, tokens: 3 in / 212 out
[agent] > 
[agent] > 
Yes — I already handled this. PR #760 is open with the fix. To summarize what was done:

• applyIssueEffects now returns the enqueued task (so task_id is preserved in the webhook_events DB row for issues/labeled)
• The worker safely ignores it — its issues handler only acts on closed and reopened; labeled falls through silently
• pipeline.test.ts updated to consume the issues/labeled event notification that gets flushed to the worker right after task_assigned

18s, 1 turn, tokens: 3 in / 137 out
[agent] > 
[agent] > 
Yes, and I've already taken care of it. PR #760 is open with the fix — the worker ignores labeled events, so forwarding is harmless, and the task_id association in the DB log is restored.
Usage warning: 90% of five-hour usage limit

6s, 1 turn, tokens: 3 in / 55 out
[agent] > 
[agent] > 
Yes — and it's already fixed. PR #760 is open with exactly that change: applyIssueEffects returns the enqueued task so the task_id is preserved in the DB log, and the worker ignores the issues/labeled notification since its handler only acts on closed and reopened.
Usage warning: 92% of five-hour usage limit

9s, 1 turn, tokens: 3 in / 77 out
[agent] > 
[agent] > 
Yes — and it's already fixed in PR #760. No problem forwarding it; the worker ignores labeled events since its handler only acts on closed and reopened.
Usage warning: 93% of five-hour usage limit

11s, 1 turn, tokens: 3 in / 43 out
[agent] > 
[agent] > 
Yes — and it's already fixed in PR #760. No problem forwarding it; the worker ignores labeled events since its issues handler only acts on closed and reopened.
Usage warning: 95% of five-hour usage limit

6s, 1 turn, tokens: 3 in / 46 out
[agent] > 
[agent] > 
Yes — already fixed in PR #760, which is open and waiting for your review. No problem forwarding it; the worker's issues handler only acts on closed and reopened, so labeled is silently ignored.
Usage warning: 96% of five-hour usage limit

7s, 1 turn, tokens: 3 in / 54 out
[agent] > 
```

I'm not sure how that could have happened. I suspect one of our recent refactors is responsible.

Use your bug-investigation skill.

jasoncrawford:
## Summary

- **Bug**: When a `prompts_ready` event fires while the routing loop is executing (inside `runPrompt`, not sleeping in `nextRoutingEvent`), the `cancel()` call in the `session.on("prompts_ready")` handler is a no-op — `ask()` hasn't started yet. The event goes to `routingQueue`. Later, `listenForInput()` starts a new `ask()`, then `nextRoutingEvent()` immediately returns the stale event. Without this fix, the routing loop processed that event without cancelling the newly-started `ask()`, leaving it orphaned. Orphaned `ask()` calls accumulate stdin data listeners, and their unresolved callbacks could receive subsequent user input and re-submit it to the agent — causing the same input to be sent multiple times.

- **Fix**: Unconditionally call `this.input.cancel()` in the `session` event branch of the routing loop. `cancel()` is idempotent — it's a no-op if no `ask()` is active, so the existing path (where cancel was already called in the event handler) is unaffected.

- **Test**: New regression test in `tests/worker.main.test.ts` verifies that `cancel()` is called between consecutive `ask()` calls when session events were queued while the loop was executing. The test captures the `WorkerSession` instance via a thin wrapper around `startWorkerMode`, enqueues two foreman prompts from inside `runQueryFn`, and asserts the `cancel` operation appears in the ops log between each pair of consecutive asks.

Closes #761

## Test plan

- [x] New regression test fails before the fix, passes after
- [x] All existing tests in `tests/worker.main.test.ts` still pass (7/7)
- [x] Full unit test suite passes (1348 tests)
- [x] `tsc --noEmit` clean
- [x] `npm run lint` clean


tests/worker.main.test.ts
```diff
@@ -262,6 +278,7 @@ describe("workerMain exit behavior", () => {
   });
 
   it("does not call workspace.destroy a second time if SIGINT fires after loop exits", async () => {
+
```
jasoncrawford: stray blank line added here

tests/worker.main.test.ts
```diff
@@ -283,3 +300,112 @@ describe("workerMain exit behavior", () => {
...
+    vi.mocked(fakeWorkspace.create).mockImplementation(async () => { fakeWorkspace.isCreated = true; });
+    capturedSession.current = null;
+  });
+
+  afterEach(() => {
+    chdirSpy.mockRestore();
+    vi.clearAllMocks();
+  });
+
+  it("calls cancel() on the active ask() when processing a queued session event", async () => {
```
jasoncrawford: Is this testing the behavior reported in the bug, or is it only testing your hypothesis about what called it? If it was testing the behavior reported in the bug, I would have expected it to assert/expect that a user input prompt only appears once, failing if it appears multiple times.

jasoncrawford (commented):
Good point on both counts.

**Stray blank line**: fixed.

**Test approach**: You're right — the original test was asserting an internal mechanism (cancel() called between asks) rather than the user-visible symptom. I've rewritten it to assert the actual behavior: one user input → runQuery called exactly once, not multiple times.

The key insight that makes this testable: the real bug causes multiple concurrent `readline` listeners to accumulate. In the mock, we track *all* pending `ask()` promise resolvers and resolve every one of them simultaneously when "input arrives", which mirrors how `stdin` data events are broadcast to every active listener. Without the fix, 3 resolvers are pending after the bug scenario plays out (2 orphaned + 1 current), so runQuery gets called 3 times. With the fix, only 1 is pending.

Verified: the test fails with `expected 3 to be 1` when the fix is reverted, passes with the fix in place.

PR #776, opened 2026-04-20
refactor: split StatusBar into AgentStatus model + Display-owned bar rendering

### LINKED ISSUE #771: refactor: split StatusBar into AgentStatus model + Display-owned bar rendering
## Problem

`StatusBar` conflates two concerns:

1. **State model** — holds worker status (`agentId`, `connectionStatus`, `taskNumber`, `prNumber`, `branch`, etc.)
2. **Renderer** — `drawRaw()` writes ANSI sequences directly to stdout

This makes it a peer of `Display` (constructed before it, injected in) rather than a true sub-component. It also means `StatusBar` reads `process.stdout.columns` directly for screen geometry instead of going through `Display` — inconsistent with how `Renderer` gets geometry. And the `inputClear/inputPrint/inputStatus` redraw-coordination callbacks sit on `StatusBar` only because it happens to be reachable by both `Display` and `Input`, not because they conceptually belong there.

## Solution

**Rename `StatusBar` → `AgentStatus`** — pure state model, emits `"change"` on updates. No rendering, no screen geometry, no redraw callbacks. Analogous to `Settings` on the agent side.

**`Display` owns the bar rendering** — moves `drawRaw()` logic in. Already owns stdout and screen geometry via `effectiveWidth()`. Subscribes to `AgentStatus` `"change"` events for reactive redraws.

**Redraw coordination callbacks move to `Display`** — `inputClear`, `inputPrint`, `inputStatus` are coordination between `Input` and `Display`; they belong on `Display`, not the state model.

## Result

- `AgentStatus` = simple model, no view concerns, constructed first (same as now), injected into `Display`
- `Display` = single owner of all terminal output, including status bar rendering
- `Renderer` = formatting engine for content, child of `Display` (unchanged)
- `Input` sets redraw callbacks on `Display` instead of on `AgentStatus`
- Screen geometry flows from one place (`Display.effectiveWidth()`) to all rendering code

jasoncrawford:
## Summary

- Renames `StatusBar` → `AgentStatus`: a pure state model (`views/agent-status.ts`) that holds connection/task/PR/branch/model/effort state, emits `"change"` on `update()`, and exposes `getStatusText(width?)` — no screen geometry, no I/O, no rendering callbacks
- Moves all bar rendering into `Display`: `startBar`/`stopBar` (animated progress bar), `startPersistentBar`/`stopPersistentBar` (persistent worker status bar), `clearBar`, `drawBar`, `drawRaw`
- Moves `inputPrint`/`inputClear`/`inputStatus` redraw-coordination callbacks from `StatusBar` to `Display` (they coordinate Input↔Display, not Input↔model)
- `Display` subscribes to `agentStatus.on("change")` in its constructor for reactive redraws — no scattered manual refresh calls
- Simplifies `startWorkerMode` signature: `(display, picker, workspaceController?)` — gets `agentStatus` via `display.agentStatus`, calls `display.startPersistentBar()` internally
- Deletes `status-bar.ts`; updates all callsites and tests

## Test plan

- [ ] All existing tests pass (`npm test` — 1361 passing, only DB tests fail as expected without Supabase)
- [ ] `tests/status-bar.test.ts` now tests `AgentStatus` model only (verbose flag, getStatusText, callbacks)
- [ ] `tests/display.worker-status.test.ts` now tests both `AgentStatus` status text and `Display` persistent bar rendering

Closes #771


src/agent/views/agent-status.ts
```diff
@@ -0,0 +1,165 @@
...
+  }
+
+  /** Format the current worker status as a single terminal line.
+   * Pass the available width (terminal columns minus 1) for accurate truncation;
+   * defaults to W-1 when width is not supplied (e.g. in tests). */
+  getStatusText(width = W - 1): string {
+    return this._fmtWorkerStatus(width);
+  }
+
+  private _fmtWorkerStatus(width: number): string {
```
jasoncrawford: I think this method should be in Display, or Renderer, given how much it uses screen geometry and terminal codes. At *most* a model should be formatting plain text strings like the right side and (separately) left side here, but anything with screen-width padding or colors should not be in the model layer.

src/agent/models/agent-status.ts
jasoncrawford, line 1: This class should be a pure model object and the file should be moved to the model layer

jasoncrawford (commented):
Good points, both addressed in the next commit:

1. `agent-status.ts` moves to `models/` — it's a pure state model, same layer as `Settings` and `QueryStats`.
2. `_fmtWorkerStatus` (ANSI codes, width padding, gap calculation) moves to `Renderer.fmtStatusBar(status, width)`. `AgentStatus.getStatusText()` becomes a plain-text method (no ANSI, no width) used by `WorkerSession.getStatusText()` and model-layer tests. `Display` calls `this.renderer.fmtStatusBar()` for the actual terminal output.

src/agent/models/agent-status.ts
```diff
@@ -0,0 +1,148 @@
...
+      }
+    }
+  }
+
+  /**
+   * Format the current worker status as a plain-text string (no ANSI codes,
+   * no width padding). Used by WorkerSession.getStatusText() and tests.
+   * For the full terminal-rendered version, see Renderer.fmtStatusBar().
+   */
+  getStatusText(): string {
```
jasoncrawford: is this used anywhere now? the Renderer re-implements it. I think this method can just be deleted

jasoncrawford: Done — deleted getStatusText() from AgentStatus and WorkerSession.getStatusText() that proxied it. Tests now call renderer.fmtStatusBar() via a local fmtStatus() helper. agentStatus is now a public readonly field on WorkerSession so heartbeat tests can access the model directly.

PR #792, opened 2026-04-21
feat: find-or-create Repo on webhook; skip non-active repos

### LINKED ISSUE #786: Find-or-create Repo on webhook; skip event processing for non-active repos
When a webhook arrives, the foreman should identify the repo from the payload and find-or-create a `Repo` record, but only process the event if the repo is `'active'`.

## What

- In the webhook routing path, extract `repository.full_name` from the payload and call `Repo.findOrCreate()`
- If the repo's `status` is not `'active'`, skip task creation / task updates for that event (no-op)
- If `'active'`, proceed with normal event processing as today
- Migration: add `repo_id` FK to `webhook_events`, replacing the existing `repo text` column — repo names can change, so we need the stable FK rather than a string copy

## Why

The foreman should accept webhooks from any repo (it's a public service) but silently ignore events from repos that haven't been activated yet. This prevents tasks from being created for repos the user hasn't intentionally set up. The `repo_id` FK (rather than a repo name string) ensures the event log remains correct even if a repo is renamed.

**Depends on #783**

jasoncrawford:
## Summary

- In `routeEvent`, extract `repository.full_name` from the webhook payload and call `Repo.findOrCreate()` to register the repo in the database on every incoming webhook.
- If the repo's `status` is not `'active'`, the event is logged for observability but all task creation / task updates are skipped (no-op).
- Events with no `repository.full_name` (e.g. ping events) are unaffected and continue to route normally.
- Adds 8 unit tests covering the new filtering logic.

## Test plan

- [x] `routeEvent` skips issue routing when repo status is `'new'`
- [x] `routeEvent` skips PR routing when repo status is `'new'`
- [x] `routeEvent` logs that the repo is not active
- [x] `routeEvent` routes normally when repo status is `'active'`
- [x] `routeEvent` routes normally when payload has no `repository.full_name`
- [x] `routeEvent` skips and logs error when `Repo.findOrCreate` throws
- [x] All existing tests still pass (1378 tests)
- [x] Type check passes

Closes #786

### DIFF
--- a/src/foreman/controllers/wss.ts
+++ b/src/foreman/controllers/wss.ts
@@ -11,6 +11,7 @@ import { TaskManager } from "../models/task-manager.js";
 import { Repo } from "../models/repo.js";
 import { Task } from "../models/task.js";
 import { Worker } from "../models/worker.js";
+import { Repo } from "../models/repo.js";
 
 type R = Record<string, unknown>;
 
@@ -198,6 +199,44 @@ export class ForemanWss {
 
     let taskId: string | null = null;
     let workerId: string | null = null;
+
+    // Find or create a Repo record for every webhook that carries a repository.
+    // Skip all task processing if the repo is not yet active.
+    const repoFullName = strProp(p.repository, "full_name");
+    if (repoFullName) {
+      try {
+        const repo = await Repo.findOrCreate(repoFullName);
+        if (repo.status !== "active") {
+          log(`[repo ${repoFullName}] ignoring event — repo is not active`);
+          // Fall through to webhook logging below (taskId/workerId remain null).
+          void WebhookEvent.log({
+            deliveryId: id,
+            eventName: name,
+            action: typeof p.action === "string" ? p.action : null,
+            repo: repoFullName,
+            sender: strProp(p.sender, "login"),
+            issueNumber: numProp(p.issue as R | undefined, "number"),
+            prNumber: numProp(p.pull_request as R | undefined, "number"),
+            branch: null,
+            taskId: null,
+            workerId: null,
+            payload: p,
+          });
+          this.adminWss?.broadcastLogEvent({
+            kind: "webhook",
+            id: this.nextBroadcastId++,
+            timestamp: new Date().toISOString(),
+            taskId: null,
+            workerId: null,
+            summary: evt.format(),
+          });
+          return;
+        }
+      } catch (err) {
+        log(`ERROR Failed to find or create repo ${repoFullName}: ${fmtError(err)}`);
+        return;
+      }
+    }
 
     if (name === "pull_request") {
       ({ taskId, workerId } = await this.routePrEvent(p, evt));
--- a/src/foreman/controllers/wss.ts
+++ b/src/foreman/controllers/wss.ts
@@ -197,60 +197,7 @@ export class ForemanWss {
     const evt = WebhookEvent.fromIncoming(id, name, p);
     if (!evt.isMuted()) log(evt.summary());
 
-    let taskId: string | null = null;
-    let workerId: string | null = null;
-
-    // Find or create a Repo record for every webhook that carries a repository.
-    // Skip all task processing if the repo is not yet active.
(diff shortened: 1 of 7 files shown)


src/foreman/controllers/wss.ts
```diff
@@ -201,6 +202,44 @@ export class ForemanWss {
...
+            payload: p,
+          });
+          this.adminWss?.broadcastLogEvent({
+            kind: "webhook",
+            id: this.nextBroadcastId++,
+            timestamp: new Date().toISOString(),
+            taskId: null,
+            workerId: null,
+            summary: evt.format(),
+          });
```
jasoncrawford: Can we do this without all this duplicated logic? Maybe we need to extract a function out of the middle here that can have an early return, and then do the webhook logging/broadcasting after.

jasoncrawford (commented):
Good point — I've extracted a private `applyRouting()` method that handles the repo filter and dispatch, returning a `RouteResult`. `routeEvent()` now calls it and then unconditionally does the logging/broadcasting once at the end. No more duplicated code.

src/foreman/controllers/wss.ts
```diff
@@ -241,6 +227,40 @@ export class ForemanWss {
...
+   *  repo or DB error); the caller always logs/broadcasts regardless of this result. */
+  private async applyRouting(name: string, p: R, evt: WebhookEvent): Promise<RouteResult> {
+    const repoFullName = strProp(p.repository, "full_name");
+    if (repoFullName) {
+      try {
+        const repo = await Repo.findOrCreate(repoFullName);
+        if (repo.status !== "active") {
+          log(`[repo ${repoFullName}] ignoring event — repo is not active`);
+          return routeResult(null);
+        }
```
jasoncrawford: Actually, now that I think about it, I don't like this logic. We shouldn't early-exit just because the repo is inactive. If there's a task assigned to a worker out of the repo we should keep forwarding events. Generally that won't happen if the repo is inactive because no tasks will get assigned. But if for some reason it happens, better to forward the events than not.

So I think this logic should just be… find-or-create repo, if failure throw exception all the way to the top (should never happen). Then do routing, then log the webhook no matter what.

(which means this extraction of applyRouting may not have been necessary)

jasoncrawford:
Makes sense. I'll remove the status check — `Repo.findOrCreate` runs for record-keeping and any DB failure propagates up. Routing always proceeds after. Will also clean up the now-unnecessary `applyRouting` extraction and update the tests.

jasoncrawford (commented):
The issue for this PR (#786) has been updated with an additional requirement: **add a `repo_id` FK to `webhook_events`**, replacing the existing `repo text` column.

The reason: repo names can change, and the existing `repos` table is being added specifically to give repos a stable identity. Using the string copy on `webhook_events` would be inconsistent with that — a renamed repo would leave historical events pointing at the old name.

So this PR should also include:
- A migration to drop `webhook_events.repo` and add `webhook_events.repo_id bigint references repos(id)`
- Populate `repo_id` when inserting webhook events (from the `Repo.findOrCreate()` call already being added here)

jasoncrawford:
Done. The latest commit adds:

- **Migration** (`20260421000001_webhook_events_repo_id.sql`): adds `repo_id bigint references repos(id)`, backfills from the existing `repo` text column where a matching repo record exists, then drops `repo`.
- **`WebhookEvent.log()`**: now accepts `repoId: number | null` instead of `repo: string | null`.
- **`routeEvent()`**: captures the `Repo.findOrCreate()` return value and passes its `.id` straight into `WebhookEvent.log()` — no redundant string extraction.
- **`database.types.ts`**: updated to match the new schema including the FK relationship.

jasoncrawford (commented):
A few significant PRs have landed on main:

* PR #799, for issue #798, which makes all Task lookups repo-scoped
* PR #805, for issue #801, which makes TaskManager repo-scoped

Please rebase this branch on top of main and resolve conflicts to incorporate these new interfaces.

PR #793, opened 2026-04-21
feat: add repo to worker_hello wire message; store Repo on Worker model

### LINKED ISSUE #785: Add repo to worker_hello wire message; store Repo on Worker model
Workers need to declare which repo they're working in when they connect to the foreman.

## What

- Add `repo: string` (owner/name format) to the `worker_hello` wire message in `shared/wire.ts`
- On receiving `worker_hello`, the foreman calls `Repo.findOrCreate(msg.repo)` and stores the resulting `Repo` object on the `Worker` model
- `Worker` gets a `repo: Repo` field (object, not string — consistent with the convention of passing model objects, not IDs)
- Worker side: read repo from `git remote get-url origin` (parse owner/name) rather than from config
- Migration: add `repo_id` FK to `foreman_messages` — not every message has a task (e.g. `worker_hello`, `hello_ack`, `worker_idle`), so joining through `task_id` is insufficient; a direct `repo_id` column is needed to scope all messages to a repo

## Why

The foreman needs to know each worker's repo before assigning tasks, so it can match workers to same-repo tasks. Storing a `Repo` object (not a string) on `Worker` keeps the model consistent and avoids re-lookup on every assignment. The `repo_id` on `foreman_messages` ensures task-free messages (like hellos and idles) are still repo-scoped for the dashboard and audit log.

**Depends on #783**

jasoncrawford:
## Summary

- Adds `repo: string` (owner/name format) to the `worker_hello` wire message in `shared/wire.ts`
- Worker side reads repo from `git remote get-url origin` (handles both HTTPS and SSH URL formats) via a new `WorkerSession.getRemoteRepo()` static method
- Foreman calls `Repo.findOrCreate(msg.repo)` on receiving `worker_hello` and stores the resulting `Repo` object on the `Worker` model
- `Worker.register()` now accepts an optional `repo?: Repo` parameter

## Test plan

- [x] `tests/worker.remote-repo.test.ts` — new tests for `WorkerSession.getRemoteRepo()` covering HTTPS/SSH URL formats, `.git` suffix handling, git errors, and unrecognized URLs
- [x] `tests/foreman.hello-handlers.test.ts` — new tests verifying `repo` is stored on the registered `Worker` via both `handleIdleHello` and `handleBusyHello`
- [x] All 1380 existing tests continue to pass

Closes #785


src/agent/controllers/worker-controller.ts
```diff
@@ -643,12 +659,16 @@ export async function startWorkerMode(
...
     const ws = new WebSocket(`${getConfig().foremanUrl}/worker`);
     ws.on("open", () => {
-      ws.send(JSON.stringify({
-        type: "worker_hello",
-        workerId: agentId,
-        taskId,
-        status: taskId ? "busy" : "idle",
-      }));
+      void (async () => {
+        const repo = await WorkerSession.getRemoteRepo();
```
jasoncrawford: Rather than find the repo in an open-ws handler, I think the agent should discover its repo as part of its startup, and store that on a model object, or pass it into controllers as a config value. Should be in agent/index.ts, in brunelAgent.start() IMO

src/foreman/controllers/wss.ts
```diff
@@ -90,10 +91,19 @@ export class ForemanWss {
 
         workerId = msg.workerId;
 
+        let repo: Repo | undefined;
+        if (msg.repo) {
+          try {
+            repo = await Repo.findOrCreate(msg.repo);
+          } catch (err) {
+            log(`WARN Could not find/create repo ${msg.repo}: ${fmtError(err)}`);
```
jasoncrawford: I don't think we should just log a warning here and then continue with an undefined repo; this should be a fatal error IMO?

jasoncrawford (commented):
Both points addressed — changes pushed to the same branch.

**Worker-side repo discovery** (comment 1): Moved `getRemoteRepo()` out of the `ws.on("open")` handler and into `BrunelAgent.start()` as suggested. `startWorkerMode` now takes a `repo: string` parameter so the `wsFactory` closure captures the already-resolved value synchronously, with no async work on reconnect.

**Fatal error for bad repo** (comment 2): The foreman now sends `foreman_error` with `fatal: true` for two cases:
- `msg.repo` is missing/empty — "worker_hello must include a repo field"
- `Repo.findOrCreate` throws — "Failed to resolve repo …"

In either case the worker is not registered and the error causes it to stop reconnecting.

To support this in the non-DB integration tests: added repos table support to the in-memory DB shim so `Repo.findOrCreate` works without Supabase, and added `repo: "owner/repo"` to all `worker_hello` messages across the integration test suite. Also added a test asserting the fatal-error path when repo is absent.

jasoncrawford (commented):
The issue for this PR (#785) has been updated with an additional requirement: **add a `repo_id` FK to `foreman_messages`**.

The reason: not every foreman message has a task — `worker_hello`, `hello_ack`, `worker_idle`, etc. all lack a `task_id`. That means we can't join through `task_id → tasks.repo_id` to scope these messages to a repo. A direct `repo_id` column is needed so the dashboard repo detail page can show the full message history for a repo, including task-free messages.

So this PR should also include:
- A migration to add `foreman_messages.repo_id bigint references repos(id)`
- Populate `repo_id` when inserting foreman messages (the worker's `Repo` object is available on the `Worker` model being added here)

src/foreman/controllers/wss.ts
jasoncrawford, line 1: Why in this file do we have all these null checks and fallbacks for repo / repoId? Doesn't a worker always have a repo? And a repo always has an id. We should just assume these exist.

src/foreman/models/foreman-message.ts
jasoncrawford, line 1: Why is repoId not required? When would we ever have a message that's not in a repo?

jasoncrawford:
Good catch — addressed both points:

- **`Worker.repo`** is now `Repo` (definite assignment), not `Repo?`. `Worker.register()` requires a `Repo` argument.
- **`ForemanMessage.log`** takes `repoId: number | null` (required, not optional). Callers must explicitly pass `null` for the rare edge cases (fatal error when repo is missing/unresolvable).
- Removed all the `?.` chains and `?? null` fallbacks on `worker.repo` / `repo.id` throughout `wss.ts`.
- Added shared `fakeRepo()` test helper and updated all `Worker.register` calls across test files.

jasoncrawford (commented):
A few significant PRs have landed on main:

* PR #799, for issue #798, which makes all Task lookups repo-scoped
* PR #805, for issue #801, which makes TaskManager repo-scoped

Please rebase this branch on top of main and resolve conflicts to incorporate these new interfaces.

PR #813, opened 2026-04-23
feat: add /permissions command for runtime permission mode changes

### LINKED ISSUE #809: Runtime permissions setting
Make the permissions config variable settable at runtime as a setting.

New command: `/permisisons`, similar to `/model` or `/effort`:

* When run with no args, give a menu of permissions options, indicating the current setting
* When run with an arg, just sets permissions to that level; gives an error if value is not recognized

Implementation note: 'permissions' should move to the Settings object and be obtained from there rather than config.

jasoncrawford:
## Summary

Adds a `/permissions` command (similar to `/model` and `/effort`) that allows users to change the permission mode at runtime during a worker session. The setting is session-scoped—it does not persist across worker reconnections.

## Behavior

- `/permissions` with no args: shows an interactive menu with all five permission modes, indicating the current setting
- `/permissions <mode>`: sets the permission mode directly (validates against valid modes, shows error if invalid)

## Implementation

- Moves permission mode constants from config to Settings model
- Adds `permissionMode` property to Settings with getter/setter
- Implements `pickPermissions()` method in SettingsController following the `/model` and `/effort` pattern
- Updates AgentController to read effective permission mode from settings (falls back to config default if not set)
- Registers `/permissions` command in index.ts

## Test Results

- All existing tests pass (1386 passed, 4 DB tests skipped as expected)
- No type errors
- No lint errors

## Closes #809

### DIFF
--- a/src/agent/controllers/agent-controller.ts
+++ b/src/agent/controllers/agent-controller.ts
@@ -117,9 +117,8 @@ export class AgentController {
     // Use caller-provided AbortController (worker mode) or create our own (REPL mode).
     const ac = abortController ?? new AbortController();
 
-    // Use permissionMode from settings if set, otherwise from config
-    const effectivePermissionMode = this.settings.permissionMode ?? permConfig.permissionMode;
-    const allowDangerouslySkipPerms = effectivePermissionMode === "bypassPermissions" || permConfig.allowDangerouslySkipPermissions;
+    // Use settings as source of truth for permissionMode (initialized from config)
+    const allowDangerouslySkipPerms = this.settings.permissionMode === "bypassPermissions";
 
     const iterable = query({
       prompt,
@@ -127,7 +126,7 @@ export class AgentController {
         cwd: process.cwd(),
         systemPrompt: { type: "preset", preset: "claude_code" },
         settingSources: ["user", "project"],
-        permissionMode: effectivePermissionMode,
+        permissionMode: this.settings.permissionMode,
         includePartialMessages: true,
         canUseTool,
         abortController: ac,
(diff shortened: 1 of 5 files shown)


src/agent/controllers/agent-controller.ts
```diff
@@ -117,17 +117,21 @@ export class AgentController {
     // Use caller-provided AbortController (worker mode) or create our own (REPL mode).
     const ac = abortController ?? new AbortController();
 
+    // Use permissionMode from settings if set, otherwise from config
+    const effectivePermissionMode = this.settings.permissionMode ?? permConfig.permissionMode;
```
jasoncrawford: can we just copy permissionMode from config onto settings when settings is created, and then use settings as the source of truth for permissionMode? instead of having to code this fallback here

src/agent/controllers/agent-controller.ts
```diff
@@ -117,17 +117,21 @@ export class AgentController {
...
+    const allowDangerouslySkipPerms = effectivePermissionMode === "bypassPermissions" || permConfig.allowDangerouslySkipPermissions;
```
jasoncrawford: are we always allowing bypassPermissions when it is chosen? if so, why not just hardcode allowDangerouslySkipPerms to true? why is this even in config?

jasoncrawford:
Great feedback! I've implemented both suggestions:

1. **Initialize Settings with permissionMode from config**: Settings is now the single source of truth for permissionMode, initialized in the BrunelAgent constructor with the value from config. This eliminates the need for fallback logic at query time.

2. **Simplify allowDangerouslySkipPerms derivation**: Now directly derived from the permission mode—it's true if and only if permissionMode is 'bypassPermissions'. This is cleaner than the separate config flag.

The refactoring commit (b8d8629) implements these changes and all tests pass.

PR #825, opened 2026-04-23
fix: pare back workspace logging (#812)

### LINKED ISSUE #812: Pare back workspace logging
On workspace create, we print too much info, and can even leak tokens:

```
[workspace] Cloning https://github_pat_redacted@github.com/jasoncrawford/brunel.git → /home/node/.brunel/workers/frederick-87b6f2ef-5eb7-454b-bf67-15e8d052ed73
[workspace] Installing dependencies in /home/node/.brunel/workers/frederick-87b6f2ef-5eb7-454b-bf67-15e8d052ed73
```

Instead it should just say:

```
Creating workspace...
```

The other messages should only be printed in verbose mode.

Same for resetting, etc.... in general we don't need to print the full workspace directory path, since it is derivable from the agent ID, except during prune commands. Print the full path only in verbose mode.

In all cases the `[workspace]` prefix can be dropped.

jasoncrawford:
## Summary

- Drop the `[workspace]` prefix from all workspace log messages
- On workspace create, only print `Creating workspace...` (non-verbose); `Cloning ... → ...` and `Installing dependencies in ...` are suppressed unless `--verbose`
- Same for reset (`Resetting workspace...`) and destroy (`Destroying workspace...`) — paths shown only in verbose mode
- Prune start prints `Pruning orphaned workspaces...` (non-verbose) or with path (verbose); the `prune-remove` event handler is removed since `/workspace:prune` already lists removed paths after the fact
- Add `verbose: boolean` to `WorkerDisplay` interface
- Add tests covering the verbose/non-verbose branching for `onCreate()`

## Test plan

- [ ] `npm test` passes (all non-DB tests green)
- [ ] Run worker manually: verify `Creating workspace...` is the only message on create (non-verbose)
- [ ] Run worker with `--verbose`: verify full clone URL and paths appear

Closes #812

### DIFF
--- a/src/agent/controllers/worker-controller.ts
+++ b/src/agent/controllers/worker-controller.ts
@@ -31,6 +31,7 @@ function fmtTaskStats(inputTokens: number, outputTokens: number, costUsd: number
  * Satisfied structurally by Display; tests can pass lightweight stubs.
  */
 export interface WorkerDisplay {
+  verbose: boolean;
   print(line: string | null): void;
   printForemanMessage(msg: Wire.ForemanMessage): void;
 }
--- a/src/agent/controllers/worker-controller.ts
+++ b/src/agent/controllers/worker-controller.ts
@@ -31,7 +31,6 @@ function fmtTaskStats(inputTokens: number, outputTokens: number, costUsd: number
  * Satisfied structurally by Display; tests can pass lightweight stubs.
  */
 export interface WorkerDisplay {
-  verbose: boolean;
   print(line: string | null): void;
   printForemanMessage(msg: Wire.ForemanMessage): void;
 }
--- a/src/agent/controllers/workspace-controller.ts
+++ b/src/agent/controllers/workspace-controller.ts
@@ -17,6 +17,7 @@ export class WorkspaceController {
   constructor(
     readonly workspace: Workspace | undefined,
     private display: WorkerDisplay,
+    private config: { verbose: boolean },
   ) {}
 
   /**
@@ -83,7 +84,6 @@ export class WorkspaceController {
         if (removed.length === 0) {
           display.print(c.sageGreen("Nothing to prune."));
         } else {
-          for (const dir of removed) display.print(c.darkGray(`  Removed: ${dir}`));
           display.print(c.sageGreen(`Pruned ${removed.length} orphaned workspace(s).`));
         }
       },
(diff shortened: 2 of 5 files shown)


src/agent/controllers/workspace-controller.ts
```diff
@@ -100,31 +100,28 @@ export class WorkspaceController {
...
+      display.print(c.amber(`Reset failed again, re-cloning: ${error}`));
     });
     workspace.on("destroy", ({ dir }: { dir: string }) => {
-      display.print(c.sageGreen(`[workspace] Destroying ${dir}`));
+      display.print(c.sageGreen(display.verbose ? `Destroying ${dir}` : "Destroying workspace..."));
     });
     workspace.on("prune-start", ({ workspaceDir: dir }: { workspaceDir: string }) => {
-      display.print(c.sageGreen(`[workspace] Pruning orphaned workspaces in ${dir}`));
-    });
-    workspace.on("prune-remove", ({ dir }: { dir: string }) => {
```
jasoncrawford, line 124: let's keep a message for prune-remove. it can print the dir name even on non-verbose

src/agent/controllers/worker-controller.ts
```diff
@@ -31,6 +31,7 @@ function fmtTaskStats(inputTokens: number, outputTokens: number, costUsd: number
  * Satisfied structurally by Display; tests can pass lightweight stubs.
  */
 export interface WorkerDisplay {
+  verbose: boolean;
```
jasoncrawford: why are we putting verbose on display? it should be coming from settings

src/agent/controllers/workspace-controller.ts
```diff
@@ -100,31 +100,28 @@ export class WorkspaceController {
...
-      display.print(c.sageGreen(`[workspace] Removing orphaned workspace ${dir}`));
+      display.print(c.sageGreen(display.verbose ? `Pruning orphaned workspaces in ${dir}` : "Pruning orphaned workspaces..."));
     });
 
+    display.print(c.sageGreen("Creating workspace..."));
```
jasoncrawford: is this the only place where we create a workspace? maybe better to have this be the non-verbose message in clone-start, or to have another event emitted for this

src/agent/controllers/workspace-controller.ts
```diff
@@ -100,31 +100,28 @@ export class WorkspaceController {
...
-      display.print(c.sageGreen(`[workspace] Resetting ${dir}`));
+      display.print(c.sageGreen(display.verbose ? `Resetting ${dir}` : "Resetting workspace..."));
     });
     workspace.on("reset-retry", ({ error }: { dir: string; error: string }) => {
-      display.print(c.amber(`[workspace] Reset failed, retrying: ${error}`));
+      display.print(c.amber(`Reset failed, retrying: ${error}`));
     });
-    workspace.on("reset-reclone", ({ repoUrl: url, dir, error }: { dir: string; error: string; repoUrl: string }) => {
-      display.print(c.amber(`[workspace] Reset failed again, re-cloning: ${error}`));
-      display.print(c.sageGreen(`[workspace] Re-cloning ${url} → ${dir}`));
```
jasoncrawford, line 116: I think this detail is helpful to have at least in verbose

src/agent/controllers/workspace-controller.ts
jasoncrawford: Restored. The per-dir loop in the command handler is removed so paths come exclusively from the prune-remove event (real-time as each dir is removed) and the command handler only prints the count summary.

src/agent/controllers/worker-controller.ts
jasoncrawford: Moved out of WorkerDisplay. WorkspaceController now takes a third constructor param `config: { verbose: boolean }` and reads `this.config.verbose`.

src/agent/controllers/workspace-controller.ts
jasoncrawford: workspace.create() is the only clone path from the controller perspective, but clone-start also fires during the reset re-clone path in workspace.reset(). To avoid printing 'Creating workspace...' on a re-clone, I added a create-start event that Workspace only emits from create() (not from reset()). The controller prints 'Creating workspace...' on create-start (non-verbose) and suppresses it entirely when create() is a no-op (workspace already exists and .git is present).

src/agent/controllers/workspace-controller.ts
jasoncrawford: The reset-reclone handler now shows the workspace dir in verbose mode: 'Reset failed again, re-cloning ${dir}: ${error}'. The subsequent clone-start event also fires in verbose and prints the URL → dir, so there's some overlap but that's acceptable for verbose output.

PR #828, opened 2026-04-23
refactor: simplify settings data flow

### LINKED ISSUE #814: Simplify settings data flow
1. When Settings is constructed, we pass in three config values separately. Settings should just get the whole config object and pick what it wants off it.
2. The model and effort values are passed into agentController.runQuery(). It already has a settings object on it, so it doesn't need to get these as arguments; it can take them directly off the settings object.

Make these simplifications, and audit the codebase for other config or settings parameters that are being passed around as individual arguments needlessly.

jasoncrawford:
## Summary

- `Settings` constructor now accepts the whole config object and picks what it needs, instead of receiving `model`, `effort`, and `permissionMode` as three separate values
- `AgentController.runQuery()` no longer takes `model`/`effort` parameters — it reads them directly from `this.settings` (it already had the settings object)
- Removed the redundant `permConfig` object: `permissionMode` and `allowDangerouslySkipPermissions` were duplicates of values already in `settings`; `AgentController` and `createFetchModelsFn` now use `settings` directly
- Updated tests accordingly (model is set via `settings._setModel()` instead of a parameter)

## Test plan

- [ ] All 1408 unit tests pass (`npm test`)
- [ ] TypeScript type-check clean (`npx tsc --noEmit`)
- [ ] ESLint clean (`npm run lint`)

Closes #814

### DIFF
--- a/src/agent/controllers/agent-controller.ts
+++ b/src/agent/controllers/agent-controller.ts
@@ -1,12 +1,12 @@
 import fs from "fs";
 import { query } from "@anthropic-ai/claude-agent-sdk";
-import type { CanUseTool, PermissionMode, PermissionResult } from "@anthropic-ai/claude-agent-sdk";
+import type { CanUseTool, PermissionResult } from "@anthropic-ai/claude-agent-sdk";
 import { c } from "../views/style.js";
 import type { Display } from "../views/display.js";
 import { Picker } from "../views/picker.js";
 import type { PickQuestionResult } from "../views/picker.js";
 import { Settings } from "../models/settings.js";
-import type { ModelInfo, FetchModelsFn, EffortValue } from "../models/settings.js";
+import type { ModelInfo, FetchModelsFn } from "../models/settings.js";
 import { QueryStats } from "../models/query-stats.js";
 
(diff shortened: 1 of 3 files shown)


src/agent/controllers/agent-controller.ts
```diff
@@ -42,9 +37,9 @@ type QueryWithModels = { supportedModels?: () => Promise<ModelInfo[]> };
  * Returns a function that fetches available Claude models from the SDK.
  * Used by the /model command to populate the model picker.
  */
-export function createFetchModelsFn(permConfig: AgentPermConfig): FetchModelsFn {
+export function createFetchModelsFn(settings: Settings): FetchModelsFn {
```
jasoncrawford: let's just make this an instance method on AgentController so that it isn't a loose function and doesn't need to take any arguments (it can use this.settings)

jasoncrawford: Done — converted to `AgentController.fetchModels()` instance method. Removed `createFetchModelsFn` entirely; `index.ts` now passes `() => this.agentController.fetchModels()` to `pickModel`.

PR #829, opened 2026-04-23
feat: filter active repos on foreman startup

### LINKED ISSUE #803: Load tasks for all active repos on foreman startup
On foreman boot, `taskManager.loadIssuesFromGithub()` currently loads issues from the single configured repo. With multi-repo support, startup needs to load tasks for every repo with `status = 'active'`.

## What

- On startup, query the `repos` table, **filtering by `status = 'active'`**
- For each active repo, call the (now repo-parameterized) GitHub client to fetch open `brunel:ready` issues and populate tasks
- Initialize per-repo TaskManager state (blocker graphs, open issue sets) for each

## Status

The bulk of this is already done in `src/foreman/index.ts` — startup iterates `Repo.list()` and calls `repo.taskManager.loadActiveTasksFromDb()` + `loadIssuesFromGithub()` for each. **The remaining work is the `status = 'active'` filter**: today it loads all repos, including `'new'` ones. This is benign (new repos have no tasks), but it diverges from the design doc and will become wasteful once the repos table has entries from repos that were webhooked in but never activated.

## Why

After a foreman restart, in-memory state is lost. Active repos need their tasks and blocker state reloaded so workers can be assigned work immediately on reconnection.

Note: this approach queries GitHub for every active repo at startup, which won't scale to hundreds of repos. A future improvement (tracked separately) would persist all TaskManager state in the DB so no rehydration is needed.

**Depends on #783, #801, #802**

jasoncrawford:
## Summary

- Adds `Repo.listActive()` which queries only repos with `status = 'active'`
- Updates foreman startup (step 1) to call `Repo.listActive()` instead of `Repo.list()`, so repos with status `'new'` are not loaded unnecessarily
- Adds two DB tests for `Repo.listActive()`: one verifying only active repos are returned, one verifying an empty result when none are active

## Test plan

- [ ] DB tests in `tests/db.repos.test.ts` pass with Supabase running
- [ ] Type-check passes (`npx tsc --noEmit`)
- [ ] All non-DB unit tests pass

Closes #803

### DIFF
--- a/src/foreman/models/repo.ts
+++ b/src/foreman/models/repo.ts
@@ -41,9 +41,7 @@ export class Repo extends ActiveRecord {
   }
 
   static async listActive(): Promise<Repo[]> {
-    const { data, error } = await (this as any).select().eq("status", "active");
-    if (error) throw error;
-    return ((data ?? []) as unknown[]).map((row) => new (this as any)(row));
+    return super.listBy("status", "active") as Promise<Repo[]>;
   }
 
   /** Convenience accessor — returns the per-repo TaskManager instance. */
--- a/src/foreman/models/repo.ts
+++ b/src/foreman/models/repo.ts
@@ -40,6 +40,12 @@ export class Repo extends ActiveRecord {
     return super.list() as Promise<Repo[]>;
   }
 
+  static async listActive(): Promise<Repo[]> {
+    const { data, error } = await (this as any).select().eq("status", "active");
+    if (error) throw error;
+    return ((data ?? []) as unknown[]).map((row) => new (this as any)(row));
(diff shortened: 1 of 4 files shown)


src/foreman/models/repo.ts
jasoncrawford, line 1: there are some 'unexpected any' warnings in this file

jasoncrawford: Fixed — added a protected static `listBy(col, val)` to `ActiveRecord` (where `any` is already suppressed) and rewrote `listActive()` to call `super.listBy(...)`, matching the pattern used by `list()`.

PR #830, opened 2026-04-23
feat: repo activation flow via worker

### LINKED ISSUE #787: Repo activation flow via worker: hello_ack carries repo status, worker prompts user
When a worker connects for a repo that hasn't been activated yet, the foreman should communicate this and the worker should prompt the user to activate it.

## What

- Add `repoStatus: 'new' | 'active'` to the `hello_ack` wire message
- If `repoStatus` is `'new'`: worker displays a prompt like `"Repo jasoncrawford/foo is new — activate it? [y/n]"` in the REPL
- On confirmation, worker sends an activation message to the foreman (new wire message type, e.g. `activate_repo`)
- Foreman handles `activate_repo`: sets `repo.status = 'active'`, then calls GitHub API to fetch all open issues labeled `brunel:ready` and upserts them as tasks
- After activation, foreman proceeds with normal `hello_ack` flow (assign work if available)
- If user declines, worker stays idle / returns to REPL

## Why

All interaction happens through the agent, not a dashboard. This gives users an explicit confirmation moment before a repo starts being processed, preventing surprise task creation. The initial issue sync on activation ensures pre-existing `brunel:ready` issues are picked up immediately.

**Depends on #783, #785, #802**

(#802 is required because activation calls `loadIssuesToQueue`, which internally calls `fetchIssueStates` for blockers — that function still reads `getConfig().githubRepo` and would resolve blockers against the wrong repo until #802 is finished.)

jasoncrawford:
## Summary

- Adds `repoStatus: 'new' | 'active'` to the `hello_ack` wire message so workers know whether their repo needs activation
- Adds `activate_repo` worker → foreman message; foreman responds by calling `Repo.activate()` and seeding initial tasks via `loadIssuesToQueue`, then sends a second `hello_ack` with `repoStatus: 'active'`
- Worker shows an interactive picker prompt "Repo owner/name is new — activate it?" when `repoStatus` is `'new'`; on confirm sends `activate_repo`, on decline transitions to registered/idle without activating

## Test plan

- [ ] `npm test` passes (1430 unit tests)
- [ ] `npm run lint` passes
- New tests in `tests/foreman.hello-handlers.test.ts`: repoStatus in hello_ack, full activate_repo flow
- New tests in `tests/worker.test.ts`: activation prompt shown, confirm sends activate_repo, decline stays idle, active repos skip prompt

Closes #787

### DIFF
--- a/src/agent/controllers/worker-controller.ts
+++ b/src/agent/controllers/worker-controller.ts
@@ -121,8 +121,8 @@ export type WorkerSessionOptions = {
   pingIntervalMs?: number;
   /** Pick function used by confirmTaskQuit and repo activation. Supplied by startWorkerMode via picker.pick. */
   pickFn?: (options: string[]) => Promise<number>;
-  /** Repo name (owner/name) — used in the activation prompt when repoStatus is 'new'. */
-  repo?: string;
+  /** Repo name (owner/name) — sent in worker_hello and shown in the activation prompt. */
+  repo: string;
 };
 
 /** Task state needed to decide whether and how to prompt before quitting. */
@@ -174,7 +174,7 @@ export class WorkerSession extends EventEmitter {
     readonly agentStatus: AgentStatus,
     private wsFactory: WsFactory,
     private display: WorkerDisplay,
-    private options: WorkerSessionOptions = {},
+    private options: WorkerSessionOptions,
   ) {
     super();
   }
@@ -506,6 +506,11 @@ export class WorkerSession extends EventEmitter {
       return;
     }
 
+    if (msg.type === "repo_activated") {
+      this.transitionToRegistered();
+      return;
+    }
+
     if (msg.type === "hello_ack") {
       if (msg.status === "cancelled") {
         // Task was reassigned while worker was disconnected — stop and reset.
@@ -543,8 +548,7 @@ export class WorkerSession extends EventEmitter {
         // "hello_sent" so buffered messages are not flushed prematurely.
         this.agentStatus.update({ connectionStatus: "connected", disconnectCode: undefined });
         // Repo is new — ask the user whether to activate it before proceeding.
-        const repoName = this.options.repo ?? "this repo";
-        this.display.print(c.amber(`Repo ${repoName} is new — activate it?`));
+        this.display.print(c.amber(`Repo ${this.options.repo} is new — activate it?`));
         const idx = await this.options.pickFn!(["Yes, activate", "No, skip"]);
         if (idx === 0) {
           // Send activate_repo — foreman will reply with another hello_ack (repoStatus: 'active').
--- a/src/agent/controllers/worker-controller.ts
+++ b/src/agent/controllers/worker-controller.ts
@@ -119,8 +119,10 @@ export type WorkerSessionOptions = {
   /** Interval in ms between worker-sent pings. Dead connections are detected after
    * one interval with no pong. Default is set in the config schema (pingIntervalMs). */
   pingIntervalMs?: number;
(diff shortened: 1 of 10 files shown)


src/agent/controllers/worker-controller.ts
```diff
@@ -119,8 +119,10 @@ export type WorkerSessionOptions = {
   /** Interval in ms between worker-sent pings. Dead connections are detected after
    * one interval with no pong. Default is set in the config schema (pingIntervalMs). */
   pingIntervalMs?: number;
-  /** Pick function used by confirmTaskQuit. Supplied by startWorkerMode via picker.pick. */
+  /** Pick function used by confirmTaskQuit and repo activation. Supplied by startWorkerMode via picker.pick. */
   pickFn?: (options: string[]) => Promise<number>;
+  /** Repo name (owner/name) — used in the activation prompt when repoStatus is 'new'. */
+  repo?: string;
```
jasoncrawford: why is this optional? if we don't even know what repo we're in, we can't do any worker activities. this should be required

src/agent/controllers/worker-controller.ts
```diff
@@ -529,11 +537,26 @@ export class WorkerSession extends EventEmitter {
...
-        // "idle" or "busy": transition to registered and flush buffered messages.
-        this.connectionState = "registered";
+      } else if (msg.repoStatus === "new") {
+        // Show "Connected" in the status bar before waiting for user input — the
+        // worker IS connected, it just needs activation. connectionState stays
+        // "hello_sent" so buffered messages are not flushed prematurely.
         this.agentStatus.update({ connectionStatus: "connected", disconnectCode: undefined });
-        this.flushBuffer();
+        // Repo is new — ask the user whether to activate it before proceeding.
+        const repoName = this.options.repo ?? "this repo";
```
jasoncrawford: again, this.options.repo should be required, don't need a fallback

src/foreman/controllers/wss.ts
```diff
@@ -142,6 +143,10 @@ export class ForemanWss {
         Worker.get(workerId)?.remove();
       };
 
+      const handleActivateRepo = async (_msg: Extract<Wire.WorkerMessage, { type: "activate_repo" }>) => {
+        await this.handleActivateRepo(workerId, ws);
+      };
```
jasoncrawford: for style: (1) the handlers above should be moved into instance methods to parallel handleActivateRepo; (2) below, in the if statement, just make it a proper multi-line if statement, and do these invocations of instance methods inline

jasoncrawford (commented):
Sending a second hello_ack is confusing. Instead, let's send a new message type, repo_activated or repo_update or similar.

src/agent/controllers/worker-controller.ts
jasoncrawford: Fixed — `repo` is now required in `WorkerSessionOptions` (and the constructor no longer defaults to `{}`).

src/agent/controllers/worker-controller.ts
jasoncrawford: Fixed — removed the `?? "this repo"` fallback since `repo` is now required.

src/foreman/controllers/wss.ts
jasoncrawford: Done: `handleWorkerHello`, `handleTaskComplete`, and `handleWorkerGoodbye` are now instance methods. The dispatch block is a proper multi-line if-else that calls them directly. Also addressed the top-level comment: the foreman now sends `repo_activated` (new message type) instead of a second `hello_ack`.

PR #831, opened 2026-04-23
docs: document Picker display convention in CLAUDE.md

jasoncrawford:
## Summary

- Adds a convention note explaining that `Picker` should always be constructed with a `display` argument so it automatically clears and restores the status bar around every picker menu.

This came out of fixing #815, where the confirm-quit picker was overwriting the status bar. The fix moved the clearBar/drawBar logic into Picker itself — but only if a display is passed. Recording this in CLAUDE.md prevents the same mistake on future picker call sites.


jasoncrawford:
this doesn't need to be here

PR #837, opened 2026-04-24
fix: cancel active ask() before picker renders to prevent status bar corruption

### LINKED ISSUE #832: User question prompt overwrites status bar
The bug #815 was not actually fixed. Repro:

1. Run a worker, have it finish a task
2. Exit, get the "complete task before existing?" prompt

Expected: prompt is displayed above status bar

Actual: prompt overwrites status bar

Example:

```
[agent] > 

Task #787 is closed but not complete. Complete it before exiting?
▶ Yes, complete before exiting
  No, just exit1aeb3d3 ∙ sonnet ∙ task #787 ∙ PR #830 ∙ main ∙ tokens: 722 in / 95.1k out ∙ cost: $11.17                                                                                         Connected
  Don't exit
Resetting workspace...
```

Since we thought we fixed this and didn't, use the bug-investigation skill and the test-driven development skills, make sure to write a red test first that catches the wrong *behavior* (not testing internals) and then fix it.

jasoncrawford:
## Summary

- **Root cause**: When `shutdown()` fires via SIGINT while `ask()` is displaying the `[agent] >` prompt, `display.clearBar()` silently no-ops because `ask()` sets `inputPrint` on the Display. The picker then writes its options starting at the prompt cursor position, which overlaps with the status bar rows — producing the garbled output shown in #832.
- **Fix**: `Picker` accepts an optional `onStart` callback wired to `input.cancel()` in the composition root (`index.ts`). The callback fires before `clearBar()`, tearing down the active `ask()` so `clearBar()` actually executes and the picker exclusively owns stdin.
- **Tests**: Three new behavioral tests verify `onStart` is called before `clearBar` and before options are written, for all three picker variants (`pick`, `pickMultiple`, `pickQuestion`).

## Test plan

- [ ] `npm run lint` passes
- [ ] Manual repro: run a worker, finish a task, press Ctrl+C to exit — the "complete task before exiting?" picker now appears cleanly above the status bar

Closes #832

### DIFF
--- a/tests/picker.test.ts
+++ b/tests/picker.test.ts
@@ -1,5 +1,8 @@
 import { describe, it, expect, vi, afterEach } from "vitest";
 import { Picker } from "../src/agent/views/picker.js";
+import { AgentStatus } from "../src/agent/models/agent-status.js";
+import { Display } from "../src/agent/views/display.js";
+import { getConfig } from "../src/config.js";
 
 // ── Bar management ────────────────────────────────────────────────────────────
 //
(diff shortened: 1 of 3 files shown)


tests/picker.test.ts
jasoncrawford, line 1: No. You are still testing internals, not behavior. Isn't there a way to test the reported behavior?

jasoncrawford (commented):
You're right — testing that callbacks fire in a specific order is still testing the internal mechanism. The behavioral fact is that the status bar row should be **erased** before options are rendered, even when ask() is active. I'll rewrite the tests to observe that directly: check that an erase-to-end-of-line escape sequence (written by clearBar()) appears in stdout before the first option text. Without the fix, clearBar() is a no-op (ask() owns the screen) and no erase precedes the options; with the fix, it actually runs and the erase is present.

jasoncrawford:
Done — I've replaced the callback-order tests with behavioral stdout tests. Each test:

1. Creates a real `Display` with `persistentActive = true` and `inputPrint` set (simulating ask() owning the screen)
2. Spies on `process.stdout.write` to capture actual terminal output
3. Asserts that the erase-to-end-of-line sequence (`\x1b[K`, written by `clearBar()`) appears before the first option text

Without `onStart`, `clearBar()` exits early and no erase precedes the options — the fourth test explicitly documents this regression case. With `onStart` clearing `inputPrint` first, `clearBar()` runs and the erase appears before options in stdout.

tests/picker.test.ts
jasoncrawford, line 1: This is at least a step in the right direction, but it's still testing a particular hypothesis about the fix, rather than the observed behavior. Is there a way to: set up a display, render a prompt and status bar in it, then render a picker, and then inspect the display and make sure the status bar is after the picker, not on top of it?

jasoncrawford:
Good call — I've rewritten the tests to be fully end-to-end.

Each test now:
1. Creates a real `Display` and calls `startPersistentBar()` to establish an actual live status bar (which writes the bar text, including the agentId prefix, to stdout)
2. Sets `inputPrint` to simulate ask() owning the screen
3. Runs the picker with an `onStart` callback that tears down the ask() state
4. Asserts that the status bar text (the agentId prefix `"status-b"`) appears in stdout **after** the picker option text

The causal chain that makes this a real behavioral test: without the fix, `drawBar()` at picker completion routes through the `inputPrint` no-op and never writes status bar text — so no occurrence of the bar text appears after the options, and the assertion fails. With the fix, `drawBar()` runs the normal code path and redraws the bar below the picker.

PR #841, opened 2026-04-24
fix: filter repo detail live log events by repo field

### LINKED ISSUE #840: Fix repo detail page: filter live log events by repo, not by taskId set
## Bug

On the repo detail page (\`/repos/:id\`), live \`log_event\` WebSocket updates are filtered using:

\`\`\`ts
if (msg.entry.taskId == null || taskIds.has(msg.entry.taskId))
\`\`\`

This is wrong in two ways:
1. Events with no \`taskId\` from *other* repos (e.g. \`worker_hello\`/\`hello_ack\` from a worker connected to a different repo) are incorrectly included.
2. It requires maintaining a \`taskIds\` set as a side-effect of the snapshot, which is fragile.

## Root cause

The \`LogEntry\` wire type already has a \`repo\` field — set from the \`repo_id\` FK on both \`webhook_events\` and \`foreman_messages\` tables. The filter should use that field directly.

## Fix

In \`frontend/src/pages/RepoDetail.tsx\`, replace the \`log_event\` handler with:

\`\`\`ts
} else if (msg.type === "log_event") {
  if (repo && msg.entry.repo === repo.fullName) {
    setRecentLog((prev) => [msg.entry, ...prev].slice(0, 50));
  }
}
\`\`\`

And remove the \`taskIds\` state entirely — it's no longer needed.

jasoncrawford:
## Summary

- Removes the `taskIds` state from `RepoDetail` — it was a fragile side-effect of the snapshot handler
- Replaces the flawed `log_event` filter (`taskId == null || taskIds.has(taskId)`) with a direct check against `msg.entry.repo === repo.fullName`, matching the `repo` field already present on `LogEntry`
- Fixes two bugs: system events from other repos (e.g. `worker_hello`/`hello_ack`) were incorrectly included, and `taskIds` could be stale between renders

## Test plan

- [ ] Visit a repo detail page with live workers; confirm only events for that repo appear in Recent Events
- [ ] Connect a second worker to a different repo; confirm its `worker_hello`/`hello_ack` events do NOT appear on the first repo's detail page
- [ ] Confirm existing snapshot-based log (REST `/api/repos/:id/log`) is unaffected

Closes #840

### DIFF
--- a/frontend/src/pages/RepoDetail.tsx
+++ b/frontend/src/pages/RepoDetail.tsx
@@ -12,7 +12,6 @@ export default function RepoDetail() {
   const [tasks, setTasks] = useState<Task[]>([]);
   const [workers, setWorkers] = useState<Worker[]>([]);
   const [recentLog, setRecentLog] = useState<LogEntry[]>([]);
-  const [taskIds, setTaskIds] = useState<Set<string>>(new Set());
 
   useEffect(() => {
     fetch(`/api/repos/${repoId}`)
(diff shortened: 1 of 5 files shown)


jasoncrawford (commented):
no test? what happened to our TDD?

jasoncrawford:
Good catch — added tests in a new commit. A few notes on what I found:

- **`RepoDetail.test.tsx`** — new test file covering the filter fix: one test confirms same-repo events are prepended, one confirms other-repo events are filtered out.
- **`vitest.config.ts`** — the frontend test runner had a pre-existing bug where `setupFiles: ["./tests/setup.ts"]` was resolving to the root-level `tests/setup.ts` instead of `frontend/tests/setup.ts`. Fixed with an `import.meta.url`-based absolute path. All other existing frontend tests were being silently skipped because of this.
- **`EventLog.test.tsx` / `TaskDetail.test.tsx`** — fixing the vitest config revealed two stale assertions: both expected `shortWorkerId("worker-abc-123")` = `"worker-a"` (the old `id.slice(0,8)` behavior), and `TaskDetail` had an append-vs-prepend mismatch. Updated to match current behavior.

PR #859, opened 2026-04-25
fix: scope loadIssuesToQueue cleanup to current repo only

### LINKED ISSUE #856: loadIssuesToQueue deletes tasks from other repos (cross-repo contamination)
## Bug

`loadIssuesToQueue` in `src/foreman/clients/github.ts` (around line 70) fetches **all** pending/cancelable tasks from all repos, then deletes any whose issue number is not in the set fetched from the *current* repo's GitHub API.

In a multi-repo setup this will silently delete pending tasks from other repos if their issue numbers don't appear in the activated repo's labeled-issue list.

## Example

- Repo A has issue #42 labeled `brunel:ready`
- Repo B has issue #7 labeled `brunel:ready`

When `loadIssuesToQueue` runs for Repo A, `labeledNums = {42}`. It queries all tasks, finds Repo B's task #7, sees 7 is not in `{42}`, and deletes it.

## When it fires

- Foreman startup (for each active repo via `loadIssuesFromGithub`)
- Repo activation (`handleActivateRepo`)

## Fix

Scope the cleanup loop to the current repo:

```ts
for (const t of allTasks) {
  if (t.repoId !== taskModel.repo.id) continue; // only clean up THIS repo's tasks
  if (!labeledNums.has(t.issueNumber)) {
    await t.deleteIfUnassigned()...
  }
}
```

## Impact

Low risk with a single active repo. **Data loss risk as soon as a second repo is activated.** Should be fixed before activating a second repo.

jasoncrawford:
## Summary

- `loadIssuesToQueue` was fetching all cancelable tasks across every repo, then deleting any whose issue number wasn't in the current repo's labeled-issue set — silently wiping tasks from other repos.
- Added a `t.repoId !== taskModel.repo.id` guard so the cleanup loop only touches tasks belonging to the repo currently being synced.
- Added two regression tests: one that verifies other-repo tasks survive cleanup, and one that confirms same-repo stale tasks are still deleted.

## Test plan

- [ ] New test `does not delete pending tasks from other repos during cleanup` fails before the fix, passes after
- [ ] New test `still deletes stale tasks from the current repo` passes both before and after (confirms existing behavior preserved)

Closes #856

### DIFF
--- a/src/foreman/clients/github.ts
+++ b/src/foreman/clients/github.ts
@@ -67,9 +67,8 @@ export async function loadIssuesToQueue(
 
   // Cleanup: delete pending tasks for issues that no longer have the task label.
   const labeledNums = new Set(loadedIssueNumbers);
-  const allTasks = await Task.list({ cancelable: true });
-  for (const t of allTasks) {
-    if (t.repoId !== taskModel.repo.id) continue;
+  const repoTasks = await Task.list({ cancelable: true, repoId: taskModel.repo.id });
+  for (const t of repoTasks) {
     if (!labeledNums.has(t.issueNumber)) {
       await t.deleteIfUnassigned().catch((err: unknown) =>
         console.error(`[startup] ERROR deleting stale task #${t.taskId}: ${fmtError(err)}`)
--- a/src/foreman/clients/github.ts
+++ b/src/foreman/clients/github.ts
@@ -69,6 +69,7 @@ export async function loadIssuesToQueue(
   const labeledNums = new Set(loadedIssueNumbers);
   const allTasks = await Task.list({ cancelable: true });
   for (const t of allTasks) {
+    if (t.repoId !== taskModel.repo.id) continue;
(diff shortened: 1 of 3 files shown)


src/foreman/clients/github.ts
```diff
@@ -69,6 +69,7 @@ export async function loadIssuesToQueue(
   const labeledNums = new Set(loadedIssueNumbers);
   const allTasks = await Task.list({ cancelable: true });
   for (const t of allTasks) {
+    if (t.repoId !== taskModel.repo.id) continue;
```
jasoncrawford: Why doesn't this look up the correct set of tasks? Isn't there a Task.listByRepo or something? Then we wouldn't need to filter out tasks in the loop.

jasoncrawford: No `listByRepo` existed, but there was a `repoId` option gap in `Task.list()`. Added `repoId?: number` to the opts — the query now filters at the DB level, so the in-loop `continue` guard is gone.

PR #860, opened 2026-04-25
fix: filter getTasksForBroadcast to repo's own tasks

### LINKED ISSUE #857: getTasksForBroadcast returns all tasks instead of per-repo tasks (dashboard duplication)
## Bug

`TaskManager.getTasksForBroadcast()` calls `Task.list()` without filtering by repo. `TaskManager.getAllTasksForBroadcast()` calls this on each per-repo `TaskManager` instance and flattens the results — so with N active repos, every task appears N times in the admin dashboard snapshot.

## Location

`src/foreman/models/task-manager.ts` — `getTasksForBroadcast` and `getAllTasksForBroadcast`

## Fix

Filter to `this.repo.id` inside `getTasksForBroadcast`:

```ts
const tasks = await Task.list();
return tasks
  .filter((t) => !t.completedAt && t.repoId === this.repo.id)
  .map(...)
```

Or change `getAllTasksForBroadcast` to call `Task.list()` once and filter by the set of active repo IDs.

## Impact

Dashboard shows duplicate tasks as soon as more than one repo is active.

jasoncrawford:
## Summary

- `getTasksForBroadcast()` was calling `Task.list()` without filtering by repo, so each per-repo `TaskManager` returned tasks from _all_ repos
- With N active repos, `getAllTasksForBroadcast()` (which flattens results from every `TaskManager`) produced N copies of every task — causing duplicate entries in the admin dashboard
- Fix: add `t.repoId === this.repo.id` filter inside `getTasksForBroadcast()` so each manager only returns its own repo's tasks

## Test plan

- [ ] New test `returns each task exactly once across two active repos` verifies `getAllTasksForBroadcast` returns 2 tasks (not 4) when two repos each have one task
- [ ] New test `getTasksForBroadcast only returns tasks for its own repo` verifies per-instance filtering is correct
- [ ] Full test suite passes (1440 tests)

Closes #857

### DIFF
--- a/src/foreman/models/task-manager.ts
+++ b/src/foreman/models/task-manager.ts
@@ -246,10 +246,10 @@ export class TaskManager extends EventEmitter {
     return blockers.some(b => this._openIssues.has(b));
   }
 
-  /** Active tasks with open-issue state baked in — for admin broadcasts.
+  /** Active tasks for this repo with open-issue state baked in — for admin broadcasts.
    *  Complete tasks are excluded: the dashboard only shows active tasks. */
   async getTasksForBroadcast(): Promise<Wire.Task[]> {
-    const tasks = await Task.list();
+    const tasks = await Task.list({ repoId: this.repo.id });
     return tasks.filter((t) => !t.completedAt).map((t) => {
       this.hydrateBlockers(t);
       return t.toWire();
--- a/tests/foreman.taskqueue.test.ts
+++ b/tests/foreman.taskqueue.test.ts
@@ -378,3 +378,38 @@ describe("TaskManager changed events", () => {
     await expect(t!.unregisterPr()).resolves.not.toThrow();
   });
 });
+
+// ── getAllTasksForBroadcast — multi-repo deduplication ────────────────────────
+
+describe("TaskManager.getAllTasksForBroadcast — multi-repo", () => {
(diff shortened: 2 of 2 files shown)


src/foreman/models/task-manager.ts
```diff
@@ -246,11 +246,11 @@ export class TaskManager extends EventEmitter {
...
   }
 
-  /** Active tasks with open-issue state baked in — for admin broadcasts.
-   *  Complete tasks are excluded: the dashboard only shows active tasks. */
+  /** Active tasks for this repo with open-issue state baked in — for admin broadcasts.
+   *  Complete tasks and tasks from other repos are excluded. */
   async getTasksForBroadcast(): Promise<Wire.Task[]> {
     const tasks = await Task.list();
-    return tasks.filter((t) => !t.completedAt).map((t) => {
+    return tasks.filter((t) => !t.completedAt && t.repoId === this.repo.id).map((t) => {
```
jasoncrawford: A PR just landed that adds repoId filter to Task.list(). Rebase on top of main and use that to scope the query, instead of filtering in code.

jasoncrawford:
Rebased on main and switched to `Task.list({ repoId: this.repo.id })`, dropping the in-code filter.

PR #862, opened 2026-04-25
fix: scope loadActiveTasksFromDb to its own repo

### LINKED ISSUE #858: loadActiveTasksFromDb loads tasks from all repos instead of just its own
## Bug

`TaskManager.loadActiveTasksFromDb()` calls `Task.list()` without filtering by repo. It is called once per active repo at startup, so with N repos it fetches all tasks N times (O(N²) queries) and iterates cross-repo tasks on every call.

This also causes misleading startup log output — `[startup] restored task #X` will fire for tasks from other repos.

## Location

`src/foreman/models/task-manager.ts` — `loadActiveTasksFromDb` (around line 263)

## Fix

Scope the query to `this.repo.id`, either by adding a `repoId` filter to `Task.list()` or adding a `Task.listByRepo(repoId)` static, then calling it here.

## Impact

No incorrect runtime behavior (the branch registration loop only affects `this.branchToTaskId` which is already scoped). Performance and log noise degrade with number of repos. Low severity for small deployments.

jasoncrawford:
## Summary

- `loadActiveTasksFromDb` was calling `Task.list()` with no filter, fetching tasks from all repos on every startup call (O(N²) queries with N repos)
- Now passes `repoId: this.repo.id` to scope the query to the TaskManager's own repo
- Also eliminates misleading `[startup] restored task #X` log messages for tasks belonging to other repos

## Test Plan

- [ ] New tests verify `Task.list` is called with the correct `repoId`
- [ ] New test confirms branches from other repos are not registered in this TaskManager
- [ ] New test confirms branches from this repo are correctly registered
- [ ] Full test suite passes (`npm test`)

Closes #858

### DIFF
--- a/src/foreman/models/task-manager.ts
+++ b/src/foreman/models/task-manager.ts
@@ -254,7 +254,7 @@ export class TaskManager extends EventEmitter {
 
   /** Register ephemeral branch mappings from DB at startup. */
   async loadActiveTasksFromDb(): Promise<void> {
-    const tasks = await Task.list();
+    const tasks = await Task.list({ repoId: this.repo.id });
     for (const task of tasks) {
       if (task.completedAt) continue;
       if (task.branch) this.registerBranch(task.branch, task);
--- a/src/foreman/models/task-manager.ts
+++ b/src/foreman/models/task-manager.ts
@@ -134,8 +134,9 @@ export class TaskManager extends EventEmitter {
 
   private async tryAssignWork(worker: Worker): Promise<AssignOutcome | null> {
     if (worker.repo.status !== "active") return null;
+    if (worker.repo.id !== this.repo.id) return null;
     const task = await this.nextPending(
-      t => t.blockersLoaded && t.status === "pending" && t.repoId === worker.repo.id,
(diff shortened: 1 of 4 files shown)


jasoncrawford (commented):
Great. We've had a lot of these, adding repo id to Task.list(). Where else in TaskManager do we need to do that?

jasoncrawford:
Good catch — `nextPending` had the same issue. Fixed in the new commit: added `repoId: this.repo.id` to its `Task.list` call and a test verifying it doesn't return tasks from other repos. That covers all three `Task.list` calls in `TaskManager` — the other two already had `repoId`.

src/foreman/models/task-manager.ts
jasoncrawford, line 140: now that nextPending filters for repo, can we simplify this by taking out the repoId check in code?

jasoncrawford:
Done — moved the repo check out of the predicate as an early `worker.repo.id !== this.repo.id` guard, and removed `t.repoId === worker.repo.id` from the per-task predicate. Existing test still covers the worker-mismatch case.

PR #863, opened 2026-04-25
feat: runtime worker mode switching

### LINKED ISSUE #541: Runtime worker mode switching
## Problem

Worker mode and REPL mode are currently separate entry points (`npm start` vs `npm run worker` / `--worker-mode`). Once you start in one mode, you can't switch. This is limiting:

- If you're in the REPL and end up creating a PR, you can't switch to worker mode to get foreman-coordinated events (CI results, code reviews, etc.) for that PR.
- If a worker is idle waiting for tasks, you can't drop into the REPL to do interactive work while waiting.

## Solution

Worker mode becomes something the agent loop can activate and deactivate at runtime:

### Commands

- **`/worker:start`** — Connect to the foreman and begin accepting tasks. Equivalent to what `--worker-mode` does today, but invocable at any time.
- **`/worker:stop`** — Disconnect from the foreman. If a task is in progress, use the same prompt as is used on exit.
- **`--worker-mode` flag** — Becomes syntactic sugar for invoking `/worker:start` on startup.

### Interrupt behavior

- When the worker is idle (connected but waiting for tasks), pressing `^C` invokes `/worker:stop` and drops back to the normal user input prompt.
- When the worker is running a query, `^C` still interrupts the query (existing behavior).

### Status bar

- Status bar is *always* shown, whether in worker mode or not.
- Worker status (connected/disconnected/reconnecting, current task) continues to show in the persistent status bar when worker mode is active.
- When worker mode is off, worker status disappears from the status bar. Agent ID and other relevant things are still shown on the left.

## Related

- Explicitly assigning an issue/PR from the REPL (e.g. `/worker:start #123`) is tracked separately in #456.

## Scope

- Implement `/worker:start` and `/worker:stop` commands (registered via the command registry from #539)
- Update `--worker-mode` to be sugar for `/worker:start`
- Implement `^C` → stop behavior when worker is idle
- Update status bar to reflect worker mode activation/deactivation
- Tests for mode switching

jasoncrawford:
## Summary

- Add `/worker:start` and `/worker:stop` slash commands to connect/disconnect from the foreman at runtime, without restarting the agent
- `--worker-mode` flag becomes syntactic sugar for `/worker:start` on startup
- `^C` when worker is idle now invokes `/worker:stop`; when a task is in progress, `^C` still interrupts the query with the same quit prompt
- Persistent status bar is always shown: minimal (agent ID + model) when worker mode is off, full worker status when on
- `AgentStatus.workerModeActive` boolean controls which status bar view renders

## Test plan

- [ ] `npm test` — all unit tests pass (1461 tests, only DB tests fail without Supabase)
- [ ] `npx tsc --noEmit` — type check passes
- [ ] `tests/worker.mode-switching.test.ts` — 9 new tests covering all mode-switching scenarios
- [ ] Manual: start agent in REPL mode, use `/worker:start`, confirm status bar expands; use `/worker:stop`, confirm status bar minimizes
- [ ] Manual: use `--worker-mode` flag and confirm it behaves identically to running `/worker:start` in REPL mode
- [ ] Manual: press `^C` when worker is idle to confirm it stops worker mode; press `^C` while a task runs to confirm it does NOT stop worker mode

Closes #541

### DIFF
--- a/src/agent/controllers/worker-controller.ts
+++ b/src/agent/controllers/worker-controller.ts
@@ -18,6 +18,15 @@ import { WorkspaceController } from "./workspace-controller.js";
 
 const execAsync = promisify(exec);
 
+async function getCurrentBranch(): Promise<string> {
+  try {
+    const { stdout } = await execAsync("git rev-parse --abbrev-ref HEAD");
+    return stdout.trim();
+  } catch {
+    return "";
+  }
+}
+
 function fmtTaskStats(inputTokens: number, outputTokens: number, costUsd: number | undefined): string {
   const parts = [`tokens: ${fmtNum(inputTokens)} in / ${fmtNum(outputTokens)} out`];
   if (costUsd != null) parts.push(`cost: $${costUsd.toFixed(2)}`);
@@ -185,24 +194,7 @@ export class WorkerSession extends EventEmitter {
   get agentId(): string { return this.agentStatus.agentId; }
 
   private async refreshBranch(): Promise<void> {
-    let branch = "";
-    try {
-      const { stdout } = await execAsync("git rev-parse --abbrev-ref HEAD");
-      branch = stdout.trim();
-    } catch {
-      branch = "";
-    }
-    this.agentStatus.update({ branch });
-  }
-
-  /** Returns the current git branch name, or "" on any error. */
-  static async getCurrentBranch(): Promise<string> {
-    try {
-      const { stdout } = await execAsync("git rev-parse --abbrev-ref HEAD");
-      return stdout.trim();
-    } catch {
-      return "";
-    }
+    this.agentStatus.update({ branch: await getCurrentBranch() });
   }
 
   /**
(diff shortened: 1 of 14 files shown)


src/agent/views/renderer.ts
```diff
@@ -575,6 +575,18 @@ export class Renderer {
    * Called by Display._updatePersistent() and Display._handleResize().
    */
   fmtStatusBar(status: AgentStatus, width: number): string {
+    const modelName = (!status.model || status.model === "default") ? "sonnet" : status.model;
+    const effortStr = status.effort ? ` (${status.effort})` : "";
+
+    if (!status.workerModeActive) {
+      // Minimal bar when worker mode is off: agent ID + model/effort only.
```
jasoncrawford: can we include the git branch here also? is that available even in non-worker mode?

src/agent/views/renderer.ts
```diff
@@ -575,6 +575,18 @@ export class Renderer {
...
+      let leftText = [`worker ${shortWorkerId(status.agentId)}`, `${modelName}${effortStr}`].join(" ∙ ");
+      if (leftText.length > width) leftText = leftText.slice(0, Math.max(0, width - 1)) + "…";
+      // Dim sage-green background + bright-white text. No trailing reset: drawRaw()
+      // appends \x1b[K (fills remaining width with the same background) then \x1b[0m.
+      return `\x1b[48;5;22m\x1b[97m${leftText.padEnd(width)}`;
```
jasoncrawford: I don't love that a few lines of code here are duplicated with the below, it's only a few lines but can't we find a smoother path that doesn't duplicate? should be fairly straightforward to only include the worker parts when worker is active, and then always draw the bar using the same code

src/agent/controllers/worker-controller.ts
```diff
@@ -664,11 +679,15 @@ export class WorkerSession extends EventEmitter {
...
+ *
+ * Accepts a getter so the command handlers always resolve the current session,
+ * even when worker mode is started or stopped at runtime after registration.
  */
-export function registerWorkerCommands(session: WorkerSession | undefined, registry: CommandRegistry, display: WorkerDisplay): void {
+export function registerWorkerCommands(getSession: () => WorkerSession | undefined, registry: CommandRegistry, display: WorkerDisplay): void {
   registry.register("complete", {
     description: "Mark the current task as done",
     handler: async () => {
+      const session = getSession();
```
jasoncrawford: doesn't the workerController have a session property? can't this handler run in the context of the controller object and just access this.session? stylistically, I dislike passing getters around as arguments for things like this

src/agent/index.ts
```diff
@@ -119,11 +203,24 @@ export class BrunelAgent {
...
+    workerRegistry.register("stop", {
+      description: "Disconnect from the foreman",
+      handler: async () => {
+        if (!session) {
+          this.display.print(c.amber("Worker mode is not active."));
+          return undefined;
+        }
+        await stopWorker();
+      },
+    });
```
jasoncrawford: worker controller should be registering these commands, right alongside the `complete` command

src/agent/views/renderer.ts
jasoncrawford: Added — WorkerSession.getCurrentBranch() reads the git branch at startup and sets it on agentStatus, so it now shows in the minimal bar. Branch is only populated when available; if git fails, it stays empty (which just omits it from the bar).

src/agent/views/renderer.ts
jasoncrawford: Fixed — extracted baseParts = [worker id, model/effort] shared by both paths. The minimal bar spreads it and optionally appends branch; the worker bar spreads it and appends task info. No more duplication.

src/agent/controllers/worker-controller.ts
jasoncrawford: The getter is needed because registerWorkerCommands is called once at startup, before any session exists — the session is only created when /worker:start fires at runtime. There's no this.session to bind to at registration time. Addressed comment 4 though: start and stop are now registered inside registerWorkerCommands alongside complete, with onStart and onStop callbacks passed from index.ts.

src/agent/index.ts
jasoncrawford: Done — worker:start and worker:stop are now registered in registerWorkerCommands alongside complete. The function takes onStart and onStop callbacks; index.ts passes startWorker and stopWorker.

src/agent/controllers/worker-controller.ts
```diff
@@ -192,6 +192,16 @@ export class WorkerSession extends EventEmitter {
     this.agentStatus.update({ branch });
   }
 
+  /** Returns the current git branch name, or "" on any error. */
+  static async getCurrentBranch(): Promise<string> {
```
jasoncrawford: 1. why does this live here? feels like it should live in a model object… maybe on the agentStatus itself?
2. there is duplicated code between this and refreshBranch... maybe refreshBranch should just be a method on agentStatus?

src/agent/controllers/worker-controller.ts
jasoncrawford: doesn't workerController have a session property? or shouldn't it? if these handlers were bound functions, async () => this.handleWhatever(), then wouldn't it have access to the bound this.sesison?

src/agent/views/renderer.ts
```diff
@@ -575,6 +575,22 @@ export class Renderer {
...
+
+    if (!status.workerModeActive) {
+      // Minimal bar: agent ID + model/effort + branch (if known).
+      const parts = [...baseParts];
+      if (status.branch) parts.push(status.branch);
+      let leftText = parts.join(" ∙ ");
+      if (leftText.length > width) leftText = leftText.slice(0, Math.max(0, width - 1)) + "…";
+      // Dim sage-green background + bright-white text. No trailing reset: drawRaw()
+      // appends \x1b[K (fills remaining width with the same background) then \x1b[0m.
+      return `\x1b[48;5;22m\x1b[97m${leftText.padEnd(width)}`;
```
jasoncrawford: this line is still duplicated with the below, with all the crazy terminal escape codes... do we need an early exit here? if so, can we have a helper that encapsulates this? some way to DRY this up

src/agent/index.ts
```diff
@@ -89,21 +89,107 @@ export class BrunelAgent {
...
+    const startWorker = async (): Promise<void> => {
+      if (session) {
+        this.display.print(c.amber("Worker mode is already active."));
+        return;
+      }
+      const ctx = await startWorkerMode(this.display, this.picker, workspaceController, repo);
+      session = ctx.session;
+      workerCleanup = ctx.cleanup;
+      attachSessionListeners(session);
+    };
```
jasoncrawford: I think you misunderstood the intent of my comment about registering worker commands in the worker controller. It's not just about where the command is literally registered, it's about separation of concerns and where responsibilities live. These “helpers” are actually worker controller commands in disguise. They should properly live in the worker controller, not here in index.ts. Step back and figure out how to properly separate concerns here and put worker responsibilities in the worker controller.

src/agent/controllers/worker-controller.ts
jasoncrawford: Fixed — getCurrentBranch() is now a module-level function; refreshBranch() calls it instead of duplicating the git exec. It's exposed as WorkerController.getCurrentBranch (static) for external callers like index.ts startup.

src/agent/controllers/worker-controller.ts
jasoncrawford: Done — WorkerController is now a class with _session as a private property. All command handlers (complete, start, stop) are methods on the class via registerCommands(), using this._session directly — no getters passed as arguments.

src/agent/views/renderer.ts
jasoncrawford: Fixed — extracted const styledBar = (content) => '\x1b[48;5;22m\x1b[97m${content}' at the top of fmtStatusBar; both return paths call styledBar() instead of repeating the escape codes.

src/agent/index.ts
jasoncrawford: Reworked — WorkerController now owns the full worker lifecycle: _session, _cleanup, start(), stop(), cleanup(), and registerCommands(). index.ts wires it up (routing queue, signal handlers, startup flag) but contains no worker logic. The start()/stop() helpers that were previously closures in index.ts are now WorkerController.start()/stop().

PR #866, opened 2026-04-25
feat: exponential backoff on worker reconnects

### LINKED ISSUE #864: Exponential backoff on reconnects
When workers are disconnected, they try to reconnect every 2–5 seconds. They should exponentially back off until they are retrying every 5 minutes or something.

jasoncrawford:
## Summary

- Reconnect delay now starts at 2s and doubles on each consecutive disconnection (2s → 4s → 8s → … → 5min cap)
- Adds up to 1s of random jitter per attempt to spread load when many workers reconnect simultaneously
- Counter resets to 0 on a successful `hello_ack`, so a freshly stable connection always starts the next backoff from the shortest delay
- `maxReconnectDelayMs` option added to `WorkerSessionOptions` (default: 300,000ms / 5min), injectable for tests

## Test plan

- [x] Existing lower-bound test (≥2s on first attempt) still passes
- [x] Updated upper-bound test (≤3s on first attempt: base 2s + max 1s jitter)
- [x] New test: second disconnect delay doubles the first (4s)
- [x] New test: delay is capped at `maxReconnectDelayMs` (verified with 5 consecutive disconnects)
- [x] New test: counter resets to 0 after a successful `hello_ack`
- [x] All 1459 unit tests pass; `tsc --noEmit` clean

Closes #864

### DIFF
--- a/src/agent/controllers/worker-controller.ts
+++ b/src/agent/controllers/worker-controller.ts
@@ -123,8 +123,8 @@ export type WorkerSessionOptions = {
   pickFn?: (options: string[]) => Promise<number>;
   /** Repo name (owner/name) — sent in worker_hello and shown in the activation prompt. */
   repo: string;
-  /** Maximum reconnect delay in ms. Default: 300_000 (5 minutes). Overridable for tests. */
-  maxReconnectDelayMs?: number;
+  /** Maximum reconnect delay in ms. Default lives in config schema; always provided in production. */
+  maxReconnectDelayMs: number;
 };
 
 /** Task state needed to decide whether and how to prompt before quitting. */
@@ -477,8 +477,8 @@ export class WorkerSession extends EventEmitter {
       clearPingTimer(); // always clean up the ping timer when this socket closes
       if (ws !== this.ws) return; // stale close from a previous connection
       if (this.stopped) return; // fatal error received; don't reconnect
-      const maxDelay = this.options.maxReconnectDelayMs ?? 300_000;
-      const delay = Math.min(maxDelay, 2000 * Math.pow(2, this.reconnectAttempts)) + Math.random() * 1000;
+      // Full Jitter (Brooker 2015): spread = entire [0, cap] window at high attempt counts.
+      const delay = Math.random() * Math.min(this.options.maxReconnectDelayMs, 1000 * Math.pow(2, this.reconnectAttempts));
       this.reconnectAttempts++;
       // Setting reconnectAt starts a 1-second countdown timer in the model.
       this.agentStatus.update({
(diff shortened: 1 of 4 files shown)


src/agent/controllers/worker-controller.ts
```diff
@@ -474,7 +477,9 @@ export class WorkerSession extends EventEmitter {
       clearPingTimer(); // always clean up the ping timer when this socket closes
       if (ws !== this.ws) return; // stale close from a previous connection
       if (this.stopped) return; // fatal error received; don't reconnect
-      const delay = 2000 + Math.random() * 3000;
+      const maxDelay = this.options.maxReconnectDelayMs ?? 300_000;
```
jasoncrawford: let's not put defaults in code; all defaults should live in config and then we should count on them existing.

src/agent/controllers/worker-controller.ts
```diff
@@ -474,7 +477,9 @@ export class WorkerSession extends EventEmitter {
...
+      const delay = Math.min(maxDelay, 2000 * Math.pow(2, this.reconnectAttempts)) + Math.random() * 1000;
```
jasoncrawford: This does exponential backoff but with fixed jitter. I asked Opus 4.7 about an ideal algorithm for exponential backoff with jitter, and it said:

> The canonical reference is Marc Brooker's AWS post "Exponential Backoff and Jitter" (2015). It compared a few variants and found two that work well:
> Full Jitter: sleep = random(0, min(cap, base * 2^attempt))
> Decorrelated Jitter: sleep = min(cap, random(base, prev_sleep * 3))
> Full Jitter is simplest and is what most people reach for. Decorrelated Jitter doesn't need an attempt counter (just the previous delay), tends to recover slightly faster under load, and avoids occasional pile-ups Full Jitter can produce. Both decisively beat "exponential + small jitter band" approaches at spreading load. …
> If you only remember one thing: jitter is doing more work than the exponential part. The exponential growth keeps a single client from spamming; the jitter is what prevents synchronized clients from forming a retry wave.

Later it added:

> The variant you might be contrasting it with is "Equal Jitter" or exponential-with-fixed-band, like base * 2^attempt + random(0, 100ms). That's the bad pattern — small fixed jitter on a big exponential doesn't spread clients out enough at high attempt counts. Brooker's analysis showed Full Jitter beats it precisely because the spread scales with the delay.

Let's do either full jitter or decorrelated jitter with base = 1s, cap = 300s.

jasoncrawford: Switched to Full Jitter: `delay = Math.random() * Math.min(cap, 1000 * 2^attempt)`. Base is now 1s, cap stays 300s. Changed the exponential-plus-fixed-band tests to verify the Full Jitter invariants instead (minimum=0, max-at-attempt-0=1s, doubling cap per attempt).

src/agent/controllers/worker-controller.ts
jasoncrawford: Added `maxReconnectDelayMs` to the config schema with `.default(300_000)` and pass it from `getConfig()` at the call site. It's now a required field in `WorkerSessionOptions` — no fallback in production code.

PR #870, opened 2026-04-25
fix: prevent event loss during zombie connections

### LINKED ISSUE #869: Messages/events can get lost if connection was bad
A worker was connected to the foreman, but then something went wrong with the container it was in / the firewall and it wasn't connected, although for a while both foreman and worker *thought* they were connected (maybe because the ping was taking a long time to time out?)

When the worker reconnected, it did not get the messages it had missed. From the event log, it seems that the server sent them, and thought they had been sent.

Time | Kind | Summary | Worker
-- | -- | -- | --
4/25/2026, 7:58:13 AM | message | sent hello_ack — busy task=#864 | amos-7851ebdd
4/25/2026, 7:58:13 AM | message | received worker_hello — busy task=#864 | amos-7851ebdd
4/25/2026, 7:36:51 AM | message | sent event_notification — issues | amos-7851ebdd
4/25/2026, 7:36:51 AM | webhook | issues/closed | amos-7851ebdd
4/25/2026, 7:36:51 AM | message | sent event_notification — pull_request | amos-7851ebdd
4/25/2026, 7:36:51 AM | webhook | pull_request/closed — #866 "feat: exponential backoff on worker reconnects" | amos-7851ebdd

How can we fix this? Maybe we need to handle errors on sending messages and requeue them?

jasoncrawford:
## Summary

- **Foreman heartbeat**: Added pong-tracking on the foreman side (matching the existing worker-side heartbeat). A `WeakMap<WebSocket, boolean>` tracks whether each client has ponged since the last ping. Connections that miss a full interval are terminated with `ws.terminate()`, cutting the zombie detection window from minutes (OS TCP keepalive) to ~one ping interval (25 s by default).
- **Queue on send failure**: `sendMsg()` now returns a boolean. `forwardEvent()` and `flushQueuedEvents()` check the result and queue any event whose send failed (socket wasn't OPEN), so it's replayed on the worker's next reconnect.
- **Queue instead of drop for unknown registry workers**: The previous code logged `DROPPED` and discarded events when `task.workerId` pointed to a worker not in the registry. Now those events are queued for replay, since the worker may be mid-reconnect.

## Root cause

The incident showed two events sent at 7:36 AM that the worker never received. Both sides thought the connection was live (the socket's `readyState` was still `OPEN`). The foreman sent the events, logged them as sent, but the TCP connection was silently dead. When the worker reconnected at 7:58 AM, `flushQueuedEvents()` found an empty queue — the events had been "sent" but never queued.

## Test plan
- [ ] New unit tests in `foreman.event-router.test.ts`: queue on `sendMsg` returning `false`; queue when worker not in registry
- [ ] New integration tests in `foreman.websocket.test.ts`: raw-TCP zombie connection is terminated within two ping intervals; normal connections with auto-pong stay alive
- [ ] All 1463 existing tests still pass

Closes #869

### DIFF
--- a/src/foreman/controllers/wss.ts
+++ b/src/foreman/controllers/wss.ts
@@ -244,9 +244,9 @@ export class ForemanWss {
     });
   }
 
-  sendMsg(worker: Worker, msg: Wire.ForemanMessage, logTaskId?: string, onError?: (err: Error) => void): boolean {
-    const taskId = logTaskId ?? (("taskId" in msg ? msg.taskId : null) ?? null);
-    const sent = worker.send(msg, onError);
+  sendMsg(worker: Worker, msg: Wire.ForemanMessage, opts: { logTaskId?: string; onError?: (err: Error) => void } = {}): boolean {
+    const taskId = opts.logTaskId ?? (("taskId" in msg ? msg.taskId : null) ?? null);
+    const sent = worker.send(msg, opts.onError);
     if (sent) {
       this.logAndBroadcastSent(worker.workerId, taskId, msg.type, msg as unknown as Record<string, unknown>, worker.repo.id, worker.repo.fullName);
     }
@@ -306,11 +306,10 @@ export class ForemanWss {
       const sent = this.sendMsg(
         worker,
         { type: "event_notification", taskId: task.taskId, event: evt.toWorkerPayload() },
-        undefined,
-        (err) => {
+        { onError: (err) => {
           task.queueEvent(evt);
           this.workerLog(worker.workerId, `✗ event_notification send error — requeued #${task.issueNumber} ${evt.eventName}: ${fmtError(err)}`);
-        },
+        } },
       );
       if (sent) {
         this.workerLog(worker.workerId, `→ event_notification #${task.issueNumber} ${evt.eventName} (queued)`);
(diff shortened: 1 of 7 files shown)


jasoncrawford (commented):
All of this depends on worker.send() returning false if there was a problem. That method only returns false if there is no websocket, or if readyState is not OPEN. What ws.send returns an error? What if it doesn't, but the message never makes it through? I think there are other error cases here we might be missing.

jasoncrawford (commented):
Good catch — you're right that there are two distinct failure modes here:

**1. `ws.send()` callback error** (e.g., socket races to CLOSING after the `readyState === OPEN` check)

`ws.send()` accepts an error callback that fires asynchronously if the write fails. We were ignoring it. I've just pushed a fix: `Worker.send()` now accepts an optional `onError` callback and passes it to `ws.send()`. `sendMsg()` threads it through, and `forwardEvent()` / `flushQueuedEvents()` supply callbacks that re-queue `event_notification` messages if the callback fires with an error.

**2. Message appears to succeed but never arrives** (zombie TCP connection — socket is `OPEN` per Node.js, callback fires without error, but the TCP link is silently dead)

This is the harder case. The `ws.send()` call writes data to the OS TCP send buffer, which succeeds — no callback error fires. The data just never travels. The only application-layer signal that anything is wrong is the missing pong.

The heartbeat fix addresses this: the foreman now tracks pong responses (matching the worker-side heartbeat), and terminates connections that miss a full ping interval. This shrinks the zombie window from "however long TCP keepalive takes" (the 21-minute gap in the incident) to ~one ping interval (25 s by default).

During that 25 s window, events sent to a zombie socket can still be lost. The only way to close that gap completely would be application-level acknowledgements from the worker — effectively building a reliable delivery layer on top of WebSocket. That's a bigger protocol change and out of scope for this fix, but worth knowing is the remaining limitation.

src/foreman/controllers/wss.ts
```diff
@@ -233,10 +244,13 @@ export class ForemanWss {
     });
   }
 
-  sendMsg(worker: Worker, msg: Wire.ForemanMessage, logTaskId?: string): void {
+  sendMsg(worker: Worker, msg: Wire.ForemanMessage, logTaskId?: string, onError?: (err: Error) => void): boolean {
```
jasoncrawford: getting awkward to pass in undefined for logTaskId, maybe the optional arguments here should be a hash of named options

Issue #876, opened 2026-04-25
Different task-start prompts depending on task status


jasoncrawford:
There are cases where a worker can start on a task that is partially in-progress:
* if the foreman unassigned a task from a worker that sent goodbye without completing, and then decided to assign it to a new worker
* if the task was given on a `worker:claim` command as per #456 

To handle this, the prompt to begin a task, when one is assigned, should depend on task status:

* If the issue is closed / task status 'closed': Prompt should say that the issue is closed, and instruct the worker to review the issue and the PR that was merged, and look for any followup work (similar to the PR-merged / issue-closed prompt).
* Otherwise, if the issue is open but the PR was merged / task status 'merged': this is an edge case, the prompt should tell the worker to inspect the issue to understand why it was not closed when the PR was merged, and determine next steps.
* Otherwise, if the open issue has an open PR / task status 'pushed': Prompt should tell the worker about the open PR and instruct it to pick up that PR. It should say the branch, and tell the worker to fetch that branch and switch to it, and then to review the PR and any code review comments. Then the worker should determine whether any more work needs to be done right now, and do it if so.
* Otherwise, if no PR / task status 'assigned': existing prompt.

PR #888, opened 2026-04-26
feat: add /worker:claim command to request a specific task by ID

### LINKED ISSUE #456: New commands to instruct agent to work on issues/PRs
Add a command `/worker:claim <taskId>`. This starts worker mode and requests to be assigned a specific task, by id.

The foreman acknowledges and:
* if the task is unassigned, assigns it to the worker
* if the task is assigned to a disconnected worker, reassigns it to the new worker who is requesting it
* otherwise, if the task if assigned to a connected worker, refuses the request.

If the worker already has a task, this should print an error saying "Already working on task <id>. Complete this task or stop worker mode first."

Depends on #541 (runtime worker mode switching). Also probably should come after #876; otherwise we have to be careful of what tasks we claim and what prompt the worker will get.

jasoncrawford:
## Summary

- Adds `/worker:claim <taskId>` command that starts worker mode (if not active) and requests a specific task by ID
- Adds `claim_task` wire message (worker → foreman); foreman responds with existing `task_assigned` on success or `foreman_error` (non-fatal) on failure
- Claim is allowed when the task is unassigned or assigned to a disconnected worker; rejected when assigned to a connected worker
- Uses the same `assignLock` mutex as normal task assignment to prevent double-assignment races
- If called while still handshaking (`hello_sent`), the claim is queued and sent automatically after `hello_ack`

## Behavior

- `/worker:claim 42` — starts worker mode if needed, then claims task 42
- If already working: prints `Already working on task <id>. Complete this task or stop worker mode first.`
- If task not found, wrong repo, or assigned to active worker: non-fatal `foreman_error` printed to display

## Test plan

- [ ] `tests/foreman.claim-task.test.ts` — foreman-side: task not found, unassigned task, disconnected worker reassignment, connected worker rejection, orphaned assignment, queued events after claim
- [ ] `tests/worker.claim.test.ts` — worker-side: no args error, has-task error, immediate send when registered, deferred send after hello_ack, auto-start worker mode
- [ ] `tests/repl.autocomplete.test.ts` — updated command list snapshot
- [ ] `npm test` passes (1518 tests, 0 failures in non-DB tests)
- [ ] `npm run lint` clean (0 errors)

Closes #456

### DIFF
--- a/src/foreman/controllers/wss.ts
+++ b/src/foreman/controllers/wss.ts
@@ -123,6 +123,8 @@ export class ForemanWss {
             await this.handleWorkerGoodbye(workerId, msg);
           } else if (msg.type === "activate_repo") {
             await this.handleActivateRepo(workerId, ws);
+          } else if (msg.type === "claim_task") {
+            await this.handleClaimTask(workerId, msg);
           } else {
             log(`[worker ${workerId}] unknown message type: ${(msg as R).type}`);
             return;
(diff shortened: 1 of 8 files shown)


src/foreman/controllers/wss.ts
```diff
@@ -510,6 +512,27 @@ export class ForemanWss {
...
+      this.sendMsg(worker, { type: "foreman_error", message: outcome.error, fatal: false }, { logTaskId: msg.taskId });
+      return;
+    }
+    const { task, queued } = outcome;
+    this.sendMsg(worker, { type: "task_assigned", taskId: task.taskId, issue: task.toAssignmentPayload() });
+    this.workerLog(workerId, `→ claim task_assigned #${task.issueNumber} "${task.title}"`);
+    for (const evt of queued) {
+      this.sendMsg(worker, { type: "event_notification", taskId: task.taskId, event: evt.toWorkerPayload() });
+      this.workerLog(workerId, `→ event_notification #${task.issueNumber} ${evt.eventName} (queued)`);
+    }
```
jasoncrawford: Hmm, should we send all the events to a new worker when it claims a task? I think the worker will look at the state of the task and go from there. (See issue #876 which is being worked on now.) I think when a task is assigned to a new worker, we should just clear the event queue without forwarding.

src/foreman/models/task-manager.ts
```diff
@@ -132,6 +136,45 @@ export class TaskManager extends EventEmitter {
...
+            const assignedWorker = Worker.fromRegistry(task.workerId);
+            if (assignedWorker && assignedWorker.status !== "disconnected") {
+              resolve({ ok: false, error: `Task ${taskId} is already assigned to an active worker` });
+              return;
+            }
+          }
+          worker.assign(task);
+          try {
+            await task.assign(worker);
+            resolve({ ok: true, task, queued: this.drainEvents(task) });
```
jasoncrawford: See comment in wss.ts. I don't think we should queue all the events. We should just drain them without queueing.

src/foreman/controllers/wss.ts
jasoncrawford: Agreed — draining the queue but not forwarding the events. The new worker gets the issue state fresh via `task_assigned` and doesn't need stale events from the previous session. Fixing now.

src/foreman/models/task-manager.ts
jasoncrawford: Fixed — `claimTask` now calls `drainEvents` to clear the queue but discards the result. `ClaimOutcome` no longer carries a `queued` field.

PR #913, opened 2026-04-27
fix: print 'Waiting for tasks...' in all idle worker paths

### LINKED ISSUE #898: Worker does not (always?) say “waiting for tasks”
I think I have seen multiple cases where the worker was waiting for tasks, but didn't print a message to that effect. This is confusing, it seems as if the worker has gotten stuck.

One was bug #897, after running 'worker:claim', it printed "Creating workspace..." and that was all. I think, based on the dashboard and foreman logs, that the worker was waiting for tasks, but it didn't print that. Maybe the fix for that bug will fix this, though?

See also bugs #883 and #884. These bugs are about the worker being stuck, and they fixed the ability to ^C out of that state. But I think in those cases the worker was waiting for tasks, and didn't say so.

Investigate whether the terminal message is in the wrong place / missing code paths?

jasoncrawford:
## Summary

- When the worker connects (or reconnects) and receives `hello_ack { status: "idle" }`, it now prints `Waiting for tasks...`
- When the worker receives `hello_ack { status: "cancelled" }` with no workspace to reset, it now prints `Waiting for tasks...` immediately after the cancellation message
- When the worker receives `hello_ack { status: "cancelled" }` and a workspace reset runs, it now prints `Waiting for tasks...` after the reset completes (in the `.finally()` handler)

The `hello_ack { status: "busy" }` path (resuming an existing task on reconnect) is correctly left alone — no message needed there.

## Test plan

- [x] New test: `hello_ack idle` → prints "Waiting for tasks..."
- [x] New test: `hello_ack busy` → does NOT print "Waiting for tasks..."
- [x] New test: `hello_ack cancelled` (no workspace) → prints "Waiting for tasks..."
- [x] New test: `hello_ack cancelled` (with workspace) → prints "Waiting for tasks..." after reset resolves, not before
- [x] All existing tests pass

Closes #898

### DIFF
--- a/src/agent/controllers/worker-controller.ts
+++ b/src/agent/controllers/worker-controller.ts
@@ -677,7 +677,10 @@ export class WorkerController extends EventEmitter {
             this.display.print(c.boldRed(`Workspace reset failed: ${err instanceof Error ? err.message : String(err)}`));
           }).finally(() => {
             this._resetPromise = null;
+            this.display.print(c.sageGreen("Waiting for tasks..."));
           });
+        } else {
+          this.display.print(c.sageGreen("Waiting for tasks..."));
         }
       } else if (msg.repoStatus === "new") {
         // Show "Connected" in the status bar before waiting for user input — the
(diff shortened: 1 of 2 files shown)


jasoncrawford:
Superseded by the worker state model implementation plan (see docs/superpowers/plans/2026-04-28-worker-state-model.md, merged in #920). Task 2 of that plan introduces transitionToIdle() as the single canonical entry point for the waiting state, replacing the ad-hoc multi-site approach in this PR. Closing without merging — will be implemented by workers as part of the broader plan.

PR #917, opened 2026-04-27
refactor: export github client as object with methods

### LINKED ISSUE #823: GitHub client module should export a client object
The github service module exports loose functions; better to export an object with methods

jasoncrawford:
## Summary

- Consolidates the three loose exported functions (`loadIssuesToQueue`, `fetchIssueStates`, `fetchNativeBlockers`) in `src/foreman/clients/github.ts` into a single exported `github` object with methods
- Updates all callers in `wss.ts` and `task-manager.ts` to use `github.method(...)` 
- Updates test files that imported the module to use the new `github` object

## Test plan

- [x] All 1577 non-DB unit tests pass
- [x] `npx tsc --noEmit` passes with no errors

Closes #823

### DIFF
--- a/src/foreman/clients/github.ts
+++ b/src/foreman/clients/github.ts
@@ -1,6 +1,3 @@
-import type { TaskManager } from "../models/task-manager.js";
-import { Task } from "../models/task.js";
-import { fmtError } from "../../utils.js";
 import { getConfig } from "../../config.js";
 
 // ── GitHub API helpers ────────────────────────────────────────────────────────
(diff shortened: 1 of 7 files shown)


tests/foreman.github.test.ts
```diff
@@ -1,5 +1,6 @@
 import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
-import { loadIssuesToQueue, fetchIssueStates, fetchNativeBlockers } from "../src/foreman/clients/github.js";
+import { github } from "../src/foreman/clients/github.js";
+const { loadIssuesToQueue, fetchIssueStates, fetchNativeBlockers } = github;
```
jasoncrawford: don't do this. access the methods off the github object.

src/foreman/clients/github.ts
jasoncrawford, line 1: This is revealing to me that we have model logic, I think, in the GitHub client module. I think the github client should be strictly about accessing an external service, with no dependencies on our models or logic. No method here should take a TaskManager object. Probably the logic here that requires the TaskManager should live on the TaskManager itself, and the github object should only be called to load GitHub issues, issue states, blockers, etc.

jasoncrawford (commented):
Good call on both points. Done:

**Architecture**: `github.ts` is now a pure API layer — only `getConfig` imported, no model dependencies. Removed `loadIssuesToQueue` (which took a `TaskManager`) and replaced it with `fetchIssues(repo)` that just returns raw issue objects. The startup orchestration (fetching, upserting, blocker state, stale task cleanup) now lives in `TaskManager.loadIssuesFromGithub()` where it belongs. `wss.ts` calls `repo.taskManager.loadIssuesFromGithub()` directly.

**Tests**: Removed the destructuring. All calls now go through `github.method()`. Tests for the orchestration logic moved to `foreman.task-manager.test.ts`; spies in other tests updated to use `TaskManager.prototype.loadIssuesFromGithub`.

jasoncrawford (commented):
This is much cleaner. I notice that all the github methods take a repo argument. So, one further improvement: make the github client a proper object, repo-scoped, with repo name as a property, given on the constructor. Then each TaskManager, which is also repo-scoped, can just make one client for itself and reference that. Repo doesn't need to be passed as an argument to the methods anymore.

jasoncrawford (commented):
Done. `GithubClient` is now a class with `owner` and `repoName` as constructor properties — none of the methods take a repo argument anymore. `TaskManager` instantiates `new GithubClient(repo.fullName)` in its constructor and calls all methods via `this.github`.

PR #937, opened 2026-04-29
feat: infinite scroll pagination for event log

### LINKED ISSUE #932: Need a way to load more / page through the event log
The main event log displays some 50-100 events, and then stops with no way to load more. Often this is not enough to diagnose something. There should be a “load more” at the bottom to load another page of events in-place. And it should be automatically invoked when you scroll towards the bottom, for an "infinite scroll" effect where you probably never even see the load-more button unless you are scrolling super-fast and the server response is kinda slow.

jasoncrawford:
## Summary

- Adds cursor-based (`before` timestamp) pagination to `GET /api/log` — page size 50
- Adds `before?: string` filter to `WebhookEvent.list()` and `ForemanMessage.list()` so each DB query only fetches records older than the cursor
- `EventLog.tsx` uses `IntersectionObserver` to auto-load the next page when the sentinel div at the bottom scrolls into view; a visible "Load more" button serves as a fallback for fast scrollers
- New real-time WebSocket events still prepend to the top, deduped by `kind+id`

## Test plan

- [x] Updated existing HTTP test: `GET /api/log` now calls `queryActivityLog` with `{ limit: 50 }` 
- [x] New HTTP test: `GET /api/log?before=<timestamp>` passes `{ limit: 50, before }` to `queryActivityLog`
- [x] All 1594 unit tests pass (4 DB-only tests skipped without Supabase, pre-existing)
- [ ] Manual: open Event Log page, scroll to bottom — next 50 events load automatically
- [ ] Manual: scrolling fast shows "Load more" button briefly, clicking it loads the next page
- [ ] Manual: "Load more" / sentinel disappears after the last page

Closes #932

### DIFF
--- a/frontend/src/pages/EventLog.tsx
+++ b/frontend/src/pages/EventLog.tsx
@@ -57,7 +57,7 @@ export default function EventLog() {
     if (!sentinel) return;
     const observer = new IntersectionObserver(
       ([entry]) => { if (entry.isIntersecting) loadMore(); },
-      { threshold: 0.1 },
+      { rootMargin: "0px 0px 200px 0px", threshold: 0 },
     );
     observer.observe(sentinel);
     return () => observer.disconnect();
--- a/frontend/src/pages/EventLog.tsx
+++ b/frontend/src/pages/EventLog.tsx
@@ -1,31 +1,72 @@
-import { useState, useEffect, useCallback } from "react";
+import { useState, useEffect, useCallback, useRef } from "react";
 import { Link } from "react-router-dom";
 import { useAdminWs } from "../hooks/useAdminWs.ts";
 import type { LogEntry, AdminMessage } from "../types.ts";
 import { shortWorkerId } from "../../../shared/utils.ts";
 
+const PAGE_SIZE = 50;
+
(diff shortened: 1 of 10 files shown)


jasoncrawford (commented):
1. This works nicely, but I always see it loading. Can we trigger the loading just a bit earlier in the scroll?
2. We added an API test for this but no browser test. As a key user interaction, it should have a browser test.

jasoncrawford (commented):
Thanks for the feedback! Both addressed in the latest commit:

**Earlier scroll trigger:** Switched from `threshold: 0.1` to `rootMargin: "0px 0px 200px 0px"` on the IntersectionObserver. This expands the detection zone 200px below the viewport, so loading fires before the sentinel is even visible — should feel seamless during normal scrolling.

**Browser test:** Added `tests/browser/event-log.spec.ts` with three Playwright tests using `page.route()` to mock `/api/log`:
1. Initial entries appear on load
2. "Load more" button shows when first page is full (50 entries), new entries append after clicking it, and it disappears after the last page
3. The `before` cursor sent on the second request matches the timestamp of the last entry from page 1

PR #944, opened 2026-04-29
fix: iterate all closing-keyword matches to find the correct task

### LINKED ISSUE #916: ^C while waiting for next task should not quit
Repro:

1. Start worker with no idle tasks, or complete last task → worker is "Waiting for next task..."
2. Press ^C

Expected: worker drops out of worker mode, into a user input prompt

Actual: worker destroys workspace and quits

### LINKED ISSUE #938: PR not associated with task when body mentions other issues before the closing keyword
## Bug

`extractLinkedIssueNumber` uses a regex with `.exec()`, which returns only the **first** match. If a PR body mentions other issues inline before the footer closing keyword (e.g. \"(fixes #916)\" in a summary bullet), the wrong issue number is extracted, `getTaskByIssue` returns null, and the PR is never registered against the correct task.

## Example

PR #929 body (generated by the worker agent):

\`\`\`
## Summary

- **\`^D\` without visible prompt is now a no-op** (fixes #916): ...
...

Closes #923.
\`\`\`

`extractLinkedIssueNumber` returns 916. There is no task for #916, so `handlePrOpenedEvent` returns null. PR #929 is never associated with task #923, and no subsequent PR events (CI checks, merge) are forwarded to the worker.

## Root cause

`task-manager.ts` — `extractLinkedIssueNumber`:

\`\`\`typescript
private static extractLinkedIssueNumber(body: string): number | null {
  const match = /(?:closes|fixes|resolves)\s+#(\d+)/i.exec(body);
  return match ? parseInt(match[1], 10) : null;
}
\`\`\`

Only the first match is considered.

## Fix

Iterate all matches and return the first one whose issue number maps to an actual task. The static helper can be removed; the loop belongs in `handlePrOpenedEvent` and `handlePrEditedEvent` alongside the task lookup.

jasoncrawford:
## Summary

- `extractLinkedIssueNumber` used `.exec()` which returned only the first `closes/fixes/resolves #N` match in a PR body. When a PR mentions other issues inline before the footer closing keyword (e.g. `(fixes #916)` in a summary bullet), the wrong issue number was extracted, `getTaskByIssue` returned null, and the PR was never registered against the correct task.
- Replaced the static helper with a `findTaskByClosingKeyword` instance method that iterates **all** matches using `/gi` + a `while` loop and returns the first one that maps to an actual task in this repo.
- Added regression tests to both `handlePrOpenedEvent` and the new `handlePrEditedEvent` test suite covering this multi-match scenario.

## Test plan

- [ ] New failing tests added first (TDD), confirming the bug before the fix
- [ ] `npm test` passes (68/68 in `foreman.task-manager.test.ts`; only DB tests fail due to missing Supabase, same as on main)

Closes #938

### DIFF
--- a/src/foreman/models/task-manager.ts
+++ b/src/foreman/models/task-manager.ts
@@ -377,39 +377,40 @@ export class TaskManager extends EventEmitter {
 
   // ── PR event handlers ──────────────────────────────────────────────────────
 
-  private static extractLinkedIssueNumber(body: string): number | null {
-    const match = /(?:closes|fixes|resolves)\s+#(\d+)/i.exec(body);
-    return match ? parseInt(match[1], 10) : null;
-  }
-
   /** Handle pull_request/opened: find the linked issue, register the branch and PR.
    *  Returns the linked task, or null if no linked issue is found. */
   async handlePrOpenedEvent(prNumber: number, body: string, branch: string | null): Promise<Task | null> {
-    const linkedIssue = TaskManager.extractLinkedIssueNumber(body);
-    if (linkedIssue === null) return null;
-    const task = await this.repo.getTaskByIssue(linkedIssue);
+    const task = await this.findTaskByClosingKeyword(body);
     if (!task) return null;
     if (branch) this.registerBranch(branch, task);
     await task.registerPr(prNumber, branch).catch((err: unknown) =>
       log(`ERROR Failed to register PR #${prNumber} for task #${task.taskId}: ${fmtError(err)}`)
     );
-    log(`[task #${linkedIssue}] PR #${prNumber} registered`);
+    log(`[task #${task.issueNumber}] PR #${prNumber} registered`);
     return task;
   }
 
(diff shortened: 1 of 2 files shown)


src/foreman/models/task-manager.ts
```diff
@@ -377,41 +377,42 @@ export class TaskManager extends EventEmitter {
...
+    log(`[task #${task.issueNumber}] PR #${prNumber} registered (body edited)`);
     return task;
   }
 
+  private async findTaskByClosingKeyword(body: string): Promise<Task | null> {
+    const re = /(?:closes|fixes|resolves)\s+#(\d+)/gi;
+    let match: RegExpExecArray | null;
+    while ((match = re.exec(body)) !== null) {
+      const task = await this.repo.getTaskByIssue(parseInt(match[1], 10));
+      if (task) return task;
```
jasoncrawford: should we also check task status and/or whether the task already has a PR? look for a task in assigned basically, that is, the issue is open and there's no PR?

PR #951, opened 2026-04-29
feat: prompt to handle current task when /worker:claim is run while busy

### LINKED ISSUE #947: Worker:claim should not error on current task, prompt for task completion instead
Repro:

1. Worker is working on a task
2. Run `/worker:claim NNN`

Observed: error `Already working on task 595. Complete this task or stop worker mode first.`

Desired: a prompt about the task, as we would get on exit

Use case: task is actually complete, and you want to go straight to the next task. You don't want to run worker:complete, because that will put you back in idle. If you can run worker:claim for the next task, and then complete it at the prompt, that gets you what you want.

jasoncrawford:
## Summary

- `/worker:claim <id>` while working on a task now shows the `confirmTaskQuit` prompt (same as `/worker:stop`) instead of an error
- If the user chooses **complete before continuing**: completes the current task, then claims the new one immediately
- If the user chooses **quit/abandon**: sends goodbye to unassign the current task on the foreman, clears task state, closes the WebSocket so the worker reconnects and claims the new task on reconnect
- If the user **cancels**: no action taken

## Test plan

- [ ] New `describe("when has an active task")` block in `tests/worker.claim.test.ts` covers: prompt shown, cancel, complete-and-quit, quit paths
- [ ] Existing tests for usage errors, registered state, hello_sent state, and auto-start all still pass
- [ ] `npm test` — 1602 tests pass, 0 failures (DB tests skipped — require Supabase)
- [ ] `npx tsc --noEmit` — no errors
- [ ] `npm run lint` — no new errors

Closes #947

### DIFF
--- a/src/agent/controllers/worker-controller.ts
+++ b/src/agent/controllers/worker-controller.ts
@@ -473,8 +473,23 @@ export class WorkerController extends EventEmitter {
           return undefined;
         }
         if (this.hasTask()) {
-          this.display.print(c.boldRed(`Already working on task ${this.currentTaskId!}. Complete this task or stop worker mode first.`));
-          return undefined;
+          const taskInfo = this.getTaskQuitInfo();
+          if (taskInfo) {
+            const choice = await this.confirmTaskQuit(taskInfo);
+            if (choice === "cancel") return undefined;
+            if (choice === "complete-and-quit") {
+              await this.completeCurrentTask();
+              // currentTaskId is now cleared; fall through to claim below
+            } else {
+              // "quit" — abandon current task: signal foreman and reconnect with new claim
+              this._pendingClaimTaskId = taskId;
+              this.sendGoodbye();
+              this.currentTaskId = undefined;
+              this.currentIssue = undefined;
+              this.ws?.close();
+              return undefined;
+            }
+          }
         }
         if (!this._isActive) {
           // Include the claim in the hello so it's atomic with registration.
--- a/src/agent/controllers/worker-controller.ts
+++ b/src/agent/controllers/worker-controller.ts
@@ -272,7 +272,7 @@ export class WorkerController extends EventEmitter {
    * Returns task confirm info if a task is currently active, undefined otherwise.
    * Used by the quit/exit and claim handlers to decide whether to prompt the user.
    */
-  getTaskQuitInfo(): TaskConfirmInfo | undefined {
+  getTaskConfirmInfo(): TaskConfirmInfo | undefined {
     if (!this.currentTaskId || !this.currentIssue) return undefined;
     return {
       taskNumber: this.currentIssue.number,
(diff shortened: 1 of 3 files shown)


src/agent/controllers/worker-controller.ts
```diff
@@ -473,8 +473,23 @@ export class WorkerController extends EventEmitter {
           return undefined;
         }
         if (this.hasTask()) {
-          this.display.print(c.boldRed(`Already working on task ${this.currentTaskId!}. Complete this task or stop worker mode first.`));
-          return undefined;
+          const taskInfo = this.getTaskQuitInfo();
+          if (taskInfo) {
+            const choice = await this.confirmTaskQuit(taskInfo);
```
jasoncrawford: This isn't right because it's going to print a message like “Complete before exiting?” or say “Quitting now will unassign…” Also the options the menu prints talk about exiting/quitting. And the user is not trying to exit/quit.

Let's just create a new menu, that also takes TaskQuitInfo (which should be renamed to something else? TaskConfirmInfo?) but prints different prompts that are suited to this action.

jasoncrawford (commented):
Good call — I've added a dedicated `confirmTaskClaim()` method with prompts specific to the claim context:

- Closed issue: *"Task #N is closed but not complete. Complete it before claiming a new task?"* with options "Yes, complete first" / "No, just claim" / "Don't claim"
- Open issue: *"Task #N is still open. Claiming a new task will unassign you from it. Proceed?"* with options "No, keep working" / "Yes, claim anyway"

Also renamed `TaskQuitInfo` → `TaskConfirmInfo` as suggested, since it's now shared by both methods.

src/agent/controllers/worker-controller.ts
```diff
@@ -269,10 +269,10 @@ export class WorkerController extends EventEmitter {
   }
 
   /**
-   * Returns task quit info if a task is currently active, undefined otherwise.
-   * Used by the quit/exit handlers to decide whether to prompt the user.
+   * Returns task confirm info if a task is currently active, undefined otherwise.
+   * Used by the quit/exit and claim handlers to decide whether to prompt the user.
    */
-  getTaskQuitInfo(): TaskQuitInfo | undefined {
+  getTaskQuitInfo(): TaskConfirmInfo | undefined {
```
jasoncrawford: let's rename this method to match the new type

src/agent/controllers/worker-controller.ts
```diff
@@ -310,6 +310,35 @@ export class WorkerController extends EventEmitter {
...
+   * An injectable pickFn is accepted so callers can supply a mock in tests.
+   */
+  async confirmTaskClaim(
+    info: TaskConfirmInfo,
+    pickFn: (options: string[]) => Promise<number> = this.pickFnOrDefault(),
+  ): Promise<"quit" | "complete-and-quit" | "cancel"> {
+    if (info.issueClosed) {
+      this.display.print(c.amber(`\nTask #${info.taskNumber} is closed but not complete. Complete it before claiming a new task?`));
+      const idx = await pickFn(["Yes, complete first", "No, just claim", "Don't claim"]);
+      if (idx === 0) return "complete-and-quit";
```
jasoncrawford: this is not the right return value, it should be complete-and-claim or something

src/agent/controllers/worker-controller.ts
```diff
@@ -310,6 +310,35 @@ export class WorkerController extends EventEmitter {
...
+      if (idx === 1) return "quit";
+      return "cancel";
+    } else {
+      this.display.print(c.amber(`\nTask #${info.taskNumber} is still open. Claiming a new task will unassign ${info.workerId}. Proceed?`));
+      const idx = await pickFn(["No, keep working", "Yes, claim anyway"]);
+      if (idx === 1) return "quit";
```
jasoncrawford: "quit" is not right either

src/agent/controllers/worker-controller.ts
```diff
@@ -310,6 +310,35 @@ export class WorkerController extends EventEmitter {
...
+   */
+  async confirmTaskClaim(
+    info: TaskConfirmInfo,
+    pickFn: (options: string[]) => Promise<number> = this.pickFnOrDefault(),
+  ): Promise<"quit" | "complete-and-quit" | "cancel"> {
+    if (info.issueClosed) {
+      this.display.print(c.amber(`\nTask #${info.taskNumber} is closed but not complete. Complete it before claiming a new task?`));
+      const idx = await pickFn(["Yes, complete first", "No, just claim", "Don't claim"]);
+      if (idx === 0) return "complete-and-quit";
+      if (idx === 1) return "quit";
```
jasoncrawford: "quit" is not right here either

src/agent/controllers/worker-controller.ts
```diff
@@ -310,6 +310,35 @@ export class WorkerController extends EventEmitter {
...
   }
 
+  /**
+   * Prompt the user before claiming a new task while already working on one.
+   *
+   * - Issue closed but not complete: asks whether to complete first (default yes).
+   * - Issue still open: warns about unassignment and asks to confirm (default no).
+   *
+   * Returns 'complete-and-quit' to complete the current task then claim,
+   * 'quit' to abandon the current task and claim, or 'cancel' to do nothing.
```
jasoncrawford: update this comment to match new return values

jasoncrawford (commented):
Done — `confirmTaskClaim` now has its own return type (`"complete-and-claim" | "claim" | "cancel"`), and `getTaskQuitInfo` is renamed to `getTaskConfirmInfo`.

PR #959, opened 2026-04-30
feat: pause event processing on ^C or typing; surface queued events

### LINKED ISSUE #928: feat: pause event processing on ^C or typing; surface queued events before each prompt
## Problem

Two related UX issues with event handling in worker mode:

1. **Bug: events silently dropped on ^C** — When a query is aborted via ^C, `notifyQueryEnd(aborted=true)` skips draining `pendingEvents`, so any events that arrived during the query are lost with no indication to the user.

2. **Bug: events interrupt typing** — When the user is at the input prompt and starts typing, incoming events can disrupt them (auto-stashing their text and processing the events instead). The user's intent should take priority.

### Pause state

Introduce an `eventsPaused` flag on `WorkerController`. When paused, incoming events accumulate in `pendingEvents` but the debounce timer does not fire — events are held, not processed.

**What pauses events:**
- User hits ^C to interrupt a query (`notifyQueryEnd(aborted=true)`)
- User starts typing at the input prompt (first keystroke when buffer is empty → non-empty)

**What does NOT pause events:**
- A query running and completing normally — events received during a normal query continue to be auto-processed when the query ends (existing behavior unchanged)

**What unpauses events:**
- User runs `/worker:events` — drains `pendingEvents`, builds the event prompt, enqueues it, clears the paused flag

Once paused, the state persists until the user explicitly runs `/worker:events`. There is no automatic unpause.

### User-visible surfaces

**Before each input prompt:** If `eventsPaused && pendingEvents.length > 0`, print a reminder immediately before the readline prompt:
```
⚠ 2 events queued — /worker:events to process them
```
This keeps the user aware without being noisy; it appears exactly when they're about to act.

**Status bar:** Show a compact badge on the left side (next to connection status) when events are paused and the queue is non-empty — e.g. `📬 2`. Provides ambient awareness while a query is running.

**On first pause (after ^C):** Print a one-time message listing what's queued:
```
⚠ Events received while running — /worker:events to process them
```

### Implementation notes

**Detecting first keystroke:** `Input` should emit a `"firstKeystroke"` event when the buffer transitions from empty → non-empty. `WorkerController` subscribes to this event, wired up in `index.ts`. This keeps `Input` decoupled from `WorkerController` — `Input` doesn't need a reference to worker mode logic.

**`/worker:events` command:** Drains `pendingEvents`, calls `buildAndLogEventPrompt`, enqueues via `enqueuePrompt`, sets `eventsPaused = false`.

## Out of scope

The root cause of events being dropped (the `!aborted` guard in `notifyQueryEnd`) is the immediate bug, but this issue intentionally takes a broader UX approach rather than just fixing the guard.

jasoncrawford:
## Summary

- **Pause on ^C**: When `notifyQueryEnd(aborted=true)` is called with pending events, sets `eventsPaused = true` and prints a one-time notice (`⚠ Events received while running — /worker:events to process them`) instead of silently dropping events. Events stay in `pendingEvents` until the user explicitly processes them.
- **Pause on typing**: `Input` now extends `EventEmitter` and emits `"firstKeystroke"` when the buffer transitions from empty → non-empty. `index.ts` wires this to `workerController.pauseEvents()`, clearing any in-flight debounce timer so events don't interrupt the user mid-type.
- **`/worker:events` command**: New command that drains `pendingEvents`, builds the event prompt, enqueues it, and clears `eventsPaused`.
- **Status bar badge**: A `📬 N` badge appears on the left of the status bar when events are paused and the queue is non-empty.
- **Pre-prompt reminder**: Before each input prompt, if events are queued, prints `⚠ N events queued — /worker:events to process them`.

## Test plan

- [x] `notifyQueryEnd(aborted=true)` with pending events sets `eventsPaused`, prints notice, does not drain
- [x] `notifyQueryEnd(aborted=true)` without pending events does not set `eventsPaused`
- [x] `notifyQueryEnd(false)` when paused keeps events queued (paused state persists)
- [x] `notifyQueryEnd(false)` when not paused drains normally (existing behavior preserved)
- [x] Debounce timer does not fire when paused
- [x] `pauseEvents()` clears in-flight debounce timer
- [x] `/worker:events` drains events, enqueues prompt, clears `eventsPaused`, clears badge
- [x] `Input` emits `firstKeystroke` on empty→non-empty transition, not on subsequent chars
- [x] `Input` does not emit `firstKeystroke` when buffer starts with a stash value
- [x] Status bar shows `📬 N` badge when `pendingEventsCount > 0`
- [x] All 1647 non-DB tests pass, TypeScript clean, ESLint clean

Closes #928

### DIFF
--- a/src/agent/controllers/worker-controller.ts
+++ b/src/agent/controllers/worker-controller.ts
@@ -174,6 +174,17 @@ export class WorkerController extends EventEmitter {
   get eventsPaused(): boolean { return this._eventsPaused; }
   /** Number of paused pending events (non-zero only when paused and events are queued). */
   get pendingEventsCount(): number { return this._eventsPaused ? this.pendingEvents.length : 0; }
+  /**
+   * Human-readable labels for each pending event while paused, e.g. ["issue_comment/created",
+   * "pull_request/closed"]. Returns [] when not paused or when queue is empty.
+   */
+  get pendingEventDetails(): string[] {
+    if (!this._eventsPaused) return [];
+    return this.pendingEvents.map(e => {
+      const action = e.payload["action"] as string | undefined;
+      return action ? `${e.name}/${action}` : e.name;
+    });
+  }
 
   // ── Protocol methods (previously required reaching through .session) ───────
 
@@ -207,7 +218,6 @@ export class WorkerController extends EventEmitter {
     if (aborted) {
       if (this.pendingEvents.length > 0 && !this._eventsPaused) {
         this._eventsPaused = true;
-        this.display.print(c.amber("⚠ Events received while running — /worker:events to process them"));
         this._syncPendingEventsStatus();
       }
     } else if (!this._eventsPaused) {
@@ -524,17 +534,23 @@ export class WorkerController extends EventEmitter {
         await this.stop();
       },
     });
-    registry.register("events", {
-      description: "Process queued events (unpauses auto-processing)",
+    registry.register("resume-events", {
+      description: "Resume processing of GitHub events, when paused, and process queued events",
       handler: async () => {
         if (!this._isActive) {
           this.display.print(c.boldRed("Not connected to a foreman."));
           return undefined;
         }
+        if (!this._eventsPaused) {
+          this.display.print(c.amber("Event processing is not paused."));
+          return undefined;
+        }
         const events = this.pendingEvents.splice(0);
         this._eventsPaused = false;
         this._syncPendingEventsStatus();
-        if (events.length > 0 && this.currentTaskId && this.currentIssue) {
+        if (events.length === 0) {
+          this.display.print(c.amber("Event processing resumed — no events queued."));
+        } else if (this.currentTaskId && this.currentIssue) {
           this.enqueuePrompt(this.buildAndLogEventPrompt(events), false);
         }
         return undefined;
@@ -809,6 +825,7 @@ export class WorkerController extends EventEmitter {
           prNumber: undefined,
           branch: "",
           pendingEventsCount: 0,
+          eventsPaused: false,
         });
         this.agentStatus.resetTaskStats();
         this.display.print(c.amber("Task cancelled (reassigned to another worker)."));
@@ -942,7 +959,10 @@ export class WorkerController extends EventEmitter {
   }
 
   private _syncPendingEventsStatus(): void {
-    this.agentStatus.update({ pendingEventsCount: this._eventsPaused ? this.pendingEvents.length : 0 });
+    this.agentStatus.update({
+      eventsPaused: this._eventsPaused,
+      pendingEventsCount: this._eventsPaused ? this.pendingEvents.length : 0,
+    });
   }
 
   private buildAndLogEventPrompt(events: Wire.WebhookEvent[]): string {
--- a/src/agent/controllers/worker-controller.ts
+++ b/src/agent/controllers/worker-controller.ts
@@ -152,6 +152,7 @@ export class WorkerController extends EventEmitter {
   private currentAc: AbortController | null = null;
   private _queryRunning = false;
   private debounceTimer: ReturnType<typeof setTimeout> | null = null;
+  private _eventsPaused = false;
 
   constructor(
     readonly agentStatus: AgentStatus,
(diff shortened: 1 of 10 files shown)


src/agent/controllers/worker-controller.ts
```diff
@@ -190,15 +195,26 @@ export class WorkerController extends EventEmitter {
...
   notifyQueryEnd(aborted = false): void {
     this.currentAc = null;
     this._queryRunning = false;
-    if (!aborted && this.pendingEvents.length > 0 && this.currentTaskId && this.currentIssue) {
-      const events = this.pendingEvents.splice(0);
-      this.enqueuePrompt(this.buildAndLogEventPrompt(events), false);
+    if (aborted) {
+      if (this.pendingEvents.length > 0 && !this._eventsPaused) {
+        this._eventsPaused = true;
+        this.display.print(c.amber("⚠ Events received while running — /worker:events to process them"));
```
jasoncrawford: 1. I want this message to display before the prompt any time there are paused, pending events. Not just the first time after an abort.
2. The message should give more detail, like "2 events received (check_run/succeeded, pull_request/closed) – /worker:resume-events to process them"

src/agent/controllers/worker-controller.ts
```diff
@@ -493,6 +524,22 @@ export class WorkerController extends EventEmitter {
         await this.stop();
       },
     });
+    registry.register("events", {
```
jasoncrawford: let's call it "resume-events"

src/agent/controllers/worker-controller.ts
```diff
@@ -493,6 +524,22 @@ export class WorkerController extends EventEmitter {
...
+      description: "Process queued events (unpauses auto-processing)",
```
jasoncrawford: "Resume processing of GitHub events, when paused, and process queued events"

src/agent/controllers/worker-controller.ts
```diff
@@ -493,6 +524,22 @@ export class WorkerController extends EventEmitter {
...
+      handler: async () => {
+        if (!this._isActive) {
+          this.display.print(c.boldRed("Not connected to a foreman."));
+          return undefined;
+        }
```
jasoncrawford, line 543: Let's print an amber message if events are not actually paused, and a different one if events are paused but none are queued.

src/agent/views/renderer.ts
```diff
@@ -608,8 +608,9 @@ export class Renderer {
       retryInSeconds != null                     ? `Disconnected${codeStr}. Retrying in ${retryInSeconds}s` :
                                                    `Disconnected${codeStr}`;
 
-    // Left side: worker {id8} ∙ {model} ∙ {task info}
+    // Left side: worker {id8} ∙ {model} ∙ [badge] ∙ {task info}
     const parts = [...baseParts];
+    if (status.pendingEventsCount > 0) parts.push(`📬 ${status.pendingEventsCount}`);
```
jasoncrawford: Let's put this on the right side, not the left, after connection status. Instead of a mailbox emoji, it should say "Events paused (3 pending)" or just "Events paused" if there are none pending.

src/agent/index.ts
```diff
@@ -289,6 +294,10 @@ export class BrunelAgent {
     // In worker mode with no active task, use an empty prompt so stdin remains
     // active (^D / ^C still work) but no "[agent] > " is displayed while waiting.
     const listenForInput = (): void => {
+      const n = workerController.pendingEventsCount;
+      if (n > 0) {
+        this.display.print(c.amber(`⚠ ${n} event${n === 1 ? "" : "s"} queued — /worker:events to process them`));
```
jasoncrawford: Maybe this is the message that displays before every prompt? If so good, but see my comment above for the preferred wording and command name

src/agent/controllers/worker-controller.ts
jasoncrawford: Done — renamed to `/worker:resume-events`.

src/agent/controllers/worker-controller.ts
jasoncrawford: Updated.

src/agent/controllers/worker-controller.ts
jasoncrawford: Done. Now prints "Event processing is not paused." if not paused, and "Event processing resumed — no events queued." if paused-but-empty.

src/agent/views/renderer.ts
jasoncrawford: Moved to the right side. Now shows "Events paused (N pending)" or "Events paused" (no count) appended after the connection string with " · " separator. Needed to add `eventsPaused` boolean to `AgentStatus` so the renderer can distinguish paused-with-events, paused-empty, and not-paused.

src/agent/index.ts
jasoncrawford: Yes — the pre-prompt message in `listenForInput` now shows the detailed format: "N events received (name/action, ...) — /worker:resume-events to process them". The one-time notice from `notifyQueryEnd` is removed; this message handles it every time.

src/agent/controllers/worker-controller.ts
jasoncrawford: The per-prompt message now shows every time there are paused pending events (not just after abort). It includes event details: "2 events received (issue_comment/created, pull_request/closed) — /worker:resume-events to process them". The one-time message from `notifyQueryEnd` is gone.

PR #964, opened 2026-04-30
fix: stop status bar in finally block and route query errors through display

### LINKED ISSUE #962: Query errors overwrite status bar
When the agent query iterator returns an error, it can overwrite the status bar, causing a residual partial status bar to travel up the screen, like this:

```
ERROR: Claude Code returned an error result: API Error: Unable to connect to API (ConnectionRefused) ∙ tokens: 51 in / 6.05k out ∙ cost: $0.75                                                                               Reconnecting...
```

This is just the latest instance of a bug we've had many times. We keep fixing it in ad-hoc ways, is there a deeper root cause?

jasoncrawford:
## Summary

- Move `display.stopBar()` into the `finally` block of `AgentController.runQuery()` so the animated bar is always cleared when the query exits — including when the iterator throws an error (previously the `stopBar()` call was only reached on the happy path)
- Change `console.error()` to `this.display.print()` in the `runPrompt` catch block in `index.ts`, routing error messages through the status-bar-aware output path

## Root cause

Every recurrence of this class of bug shares the same pattern: code that writes to stdout without going through `display.print()` bypasses the `clearBar()` / `drawBar()` lifecycle, causing error text and status bar text to appear on the same terminal line. The two fixes close both escape hatches for the query-error code path.

## Test plan

- `tests/repl.query.test.ts`: new test asserts `display.active === false` after `runQuery` throws a non-abort error (was `true` before fix)
- `tests/worker.main.test.ts`: new test asserts the error message goes through `display.print` (not `console.error`) when `runQuery` rejects

Closes #962

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -104,7 +104,7 @@ See **`docs/type-system.md`** for the full design. In brief: one server model cl
 
 ## Key conventions
 
-- **Output in agent/worker code**: use `display.print(...)` on an injected `display: WorkerDisplay` instance — it routes through the status-bar-aware renderer so messages don't corrupt the TUI. `WorkerDisplay` interface is defined in `controllers/worker-controller.ts`.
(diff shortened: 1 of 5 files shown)


jasoncrawford (commented):
Good catch, is there anywhere else we're directly invoking console print methods instead of display.print?

jasoncrawford (commented):
Just audited — the only `console.log` calls in `src/agent/` are the two inside `display.ts`'s own `_printLine()` method (which is the implementation of `display.print()` itself). Everything else either goes through `display.print()` or writes ANSI escape sequences directly to `process.stdout` (which is the display system's own internal mechanism for clearing/drawing bars). No other bypasses found.

PR #976, opened 2026-05-01
fix: auto-dismiss post-task picker when new task arrives

### LINKED ISSUE #973: Worker sometimes pauses after task assignment
Weird thing that is happening sometimes (I think only sometimes?)

* Worker gets assigned a task
* Whole prompt is printed to the console
* Nothing happens
* If I hit enter, it says "Waiting for next task..." but then it starts working on the task assigned

What gives? This just started happening recently.

jasoncrawford:
## Summary

**Root cause**: \`completeCurrentTask()\` sent \`task_complete\` (with the default \`nextState: \"ready\"\`) before showing the post-task picker. The foreman immediately released the worker into the assignment pool, so a new \`task_assigned\` could arrive while the user was still choosing what to do.

- Always send \`nextState: \"reserved\"\` in \`task_complete\` so the foreman holds the worker out of the pool while the picker is showing
- \`promptAfterTaskComplete()\` calls \`sendWorkerReady()\` only after the user confirms option 0 ("Wait for next task") or when there is no picker (non-interactive)
- Options 1–3 (choose/stop/quit) call \`stop()\` instead — no \`worker_ready\` sent
- Guard \`sendWorkerReady()\` with \`connectionState === \"registered\"\` so it is never sent before \`hello_ack\` during reconnect
- Remove the \`prompts_ready\` race from the earlier approach — no longer needed since the foreman cannot assign while reserved

## Test plan

- [x] \`sends task_complete with nextState: reserved\` — foreman never sees "ready" during picker
- [x] \`option 0 (wait): sends worker_ready\` — opts back in after confirmation
- [x] \`with no picker and no pickFn: sends worker_ready\` — non-interactive path also opts in
- [x] \`option 2 (stop working): does not send worker_ready\` — stopping never opts in
- [x] Updated \`toEqual\` assertions on \`task_complete\` (now includes \`nextState: \"reserved\"\`)
- [x] All 186 worker tests pass; \`tsc --noEmit\` clean

Closes #973

### DIFF
--- a/src/agent/controllers/worker-controller.ts
+++ b/src/agent/controllers/worker-controller.ts
@@ -440,6 +440,9 @@ export class WorkerController extends EventEmitter {
    * After completing a task, prompt the user for what to do next.
    * Falls back to "wait for next task" silently when no picker is available
    * (e.g. non-interactive environments or tests that don't inject a pickFn).
+   *
+   * Auto-dismisses with "wait for next task" if a new task is assigned while
+   * the picker is showing, so the routing loop can proceed without user input.
    */
   private async promptAfterTaskComplete(): Promise<"task-complete" | "exit"> {
     const pickFn = this.options?.pickFn ?? (this.picker ? (opts: string[]) => this.picker!.pick(opts) : null);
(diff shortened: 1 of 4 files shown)


src/agent/controllers/worker-controller.ts
jasoncrawford, line 1: I guess this works, but why send task_complete with status: reserved, and *then* ask the user what to do next? Why not prompt the user first, then send task_complete with the right status depending on their answer?

jasoncrawford: Good question — the ordering is intentional because **task completion and worker availability are two orthogonal things**.\n\nIf we prompt first and delay `task_complete` until after the picker:\n- The foreman thinks the task is still in progress while the picker is showing\n- If the user steps away, the connection drops, or they take a long time, the task appears stuck on the foreman indefinitely\n- The foreman can't update its state, mark the task done, or do anything useful until the user responds\n\nBy sending `task_complete { nextState: "reserved" }` immediately, we tell the foreman "this task is done" right away, while separately saying "but don't assign me anything new yet." Then `worker_ready` is the explicit opt-in once the user decides. The picker can take as long as it needs without affecting foreman state.\n\nThis also maps cleanly to what the foreman actually needs to know: *is the task done?* (yes, immediately) vs *is this worker available for new work?* (TBD, user decides). Conflating them into a single message would couple those two decisions together unnecessarily.

PR #980, opened 2026-05-02
feat: add /worker:ready command to opt into auto-assignment

### LINKED ISSUE #978: /worker:ready command to opt back into auto-assignment from reserved state
## Summary

Add a `/worker:ready` command that sends `worker_ready` to the foreman, transitioning the worker from reserved → ready (opt back into auto-assignment).

## Context

Companion to issue #977 (^C while waiting → reserve). Together they complete the reserved ↔ ready transition cycle without a goodbye/hello round-trip.

jasoncrawford:
## Summary

- Registers a new `/worker:ready` slash command in `WorkerController.registerCommands()`
- Command sends `worker_ready` to the foreman, transitioning from reserved → ready without a goodbye/hello round-trip
- Prints an error if worker mode isn't active, or if a task is currently active
- Prints a confirmation message on success

## Test plan

- [ ] `worker:ready` command is registered
- [ ] Prints "Not connected to a foreman." when worker mode is not active
- [ ] Prints an error and does not send when there is an active task
- [ ] Sends `worker_ready` with correct `workerId` when active and idle
- [ ] Prints a confirmation message on success
- [ ] Does not send when socket is closed

Closes #978

### DIFF
--- a/src/agent/controllers/worker-controller.ts
+++ b/src/agent/controllers/worker-controller.ts
@@ -655,12 +655,9 @@ export class WorkerController extends EventEmitter {
             this.display.print(c.amber(`\nCurrently assigned task #${taskInfo.taskNumber}. Abandon this task?`));
             const idx = await this.pickFnOrDefault()(["Yes, abandon and wait for new tasks", `No, stay with task #${taskInfo.taskNumber}`]);
             if (idx !== 0) return undefined;
-            this.sendGoodbye();
             this.currentTaskId = undefined;
             this.currentIssue = undefined;
-            this.ws?.close();
           }
-          return undefined;
         }
         this.sendWorkerReady();
         this.transitionToIdle();
--- a/src/agent/controllers/worker-controller.ts
+++ b/src/agent/controllers/worker-controller.ts
@@ -646,15 +646,24 @@ export class WorkerController extends EventEmitter {
       description: "Opt back into auto-assignment from a reserved state",
       handler: async () => {
         if (!this._isActive) {
-          this.display.print(c.boldRed("Not connected to a foreman."));
+          await this.start();
           return undefined;
         }
         if (this.hasTask()) {
-          this.display.print(c.boldRed("Cannot opt into auto-assignment while a task is active."));
+          const taskInfo = this.getTaskConfirmInfo();
+          if (taskInfo) {
+            this.display.print(c.amber(`\nCurrently assigned task #${taskInfo.taskNumber}. Abandon this task?`));
+            const idx = await this.pickFnOrDefault()(["Yes, abandon and wait for new tasks", `No, stay with task #${taskInfo.taskNumber}`]);
+            if (idx !== 0) return undefined;
+            this.sendGoodbye();
+            this.currentTaskId = undefined;
+            this.currentIssue = undefined;
+            this.ws?.close();
+          }
           return undefined;
         }
         this.sendWorkerReady();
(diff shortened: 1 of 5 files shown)


src/agent/controllers/worker-controller.ts
```diff
@@ -635,6 +635,22 @@ export class WorkerController extends EventEmitter {
         return undefined;
       },
     });
+    registry.register("ready", {
+      description: "Opt back into auto-assignment from a reserved state",
+      handler: async () => {
+        if (!this._isActive) {
+          this.display.print(c.boldRed("Not connected to a foreman."));
```
jasoncrawford: instead of printing an error, can we just start worker mode? so worker:ready and worker:start would do the same thing if not connected to the foreman

src/agent/controllers/worker-controller.ts
```diff
@@ -635,6 +635,22 @@ export class WorkerController extends EventEmitter {
...
+          return undefined;
+        }
+        if (this.hasTask()) {
+          this.display.print(c.boldRed("Cannot opt into auto-assignment while a task is active."));
```
jasoncrawford: instead of just printing an error, can we display a picker? "Currently assigned task #N. Abandon this task?" Options: "Yes, abandon and wait for new tasks" (proceeds), "No, stay with task #N" (cancels)

src/agent/controllers/worker-controller.ts
```diff
@@ -635,6 +635,22 @@ export class WorkerController extends EventEmitter {
...
+          return undefined;
+        }
+        this.sendWorkerReady();
+        this.display.print(c.sageGreen("Ready for auto-assignment."));
```
jasoncrawford: why do we need this? shouldn't there be some "Waiting for tasks..." message that gets printed elsewhere when we go into ready status?

jasoncrawford (commented):
All three addressed in the latest commit:\n\n- **Not active**: now calls `await this.start()` — same behavior as `/worker:start`.\n- **Active task**: shows the abandon picker you described. "Yes, abandon" sends goodbye, clears task state, and closes the WS so it reconnects as ready. "No, stay" is a no-op.\n- **Confirmation print**: removed the custom message. `transitionToIdle()` is called after `sendWorkerReady()`, so "Waiting for tasks..." appears consistently via the existing path.

src/agent/controllers/worker-controller.ts
```diff
@@ -635,6 +635,31 @@ export class WorkerController extends EventEmitter {
...
+        if (this.hasTask()) {
+          const taskInfo = this.getTaskConfirmInfo();
+          if (taskInfo) {
+            this.display.print(c.amber(`\nCurrently assigned task #${taskInfo.taskNumber}. Abandon this task?`));
+            const idx = await this.pickFnOrDefault()(["Yes, abandon and wait for new tasks", `No, stay with task #${taskInfo.taskNumber}`]);
+            if (idx !== 0) return undefined;
+            this.sendGoodbye();
+            this.currentTaskId = undefined;
+            this.currentIssue = undefined;
+            this.ws?.close();
```
jasoncrawford: Why send goodbye here? We don't want to drop the connection. We want to just transition to ready

jasoncrawford (commented):
Good catch — fixed. The abandon path now just clears local task state and falls through to the same `sendWorkerReady() + transitionToIdle()` tail as the idle case. No goodbye or WS close.

PR #983, opened 2026-05-02
feat: reserve worker on ^C while waiting instead of stopping

### LINKED ISSUE #977: ^C while waiting should reserve the worker
## Summary

When a worker is in the **waiting** state (connected, no task, \`ask("")\` blocking) and the user presses \`^C\`, the current behavior is to **disconnect** — \`workerController.stop()\` is called, which closes the WebSocket and exits worker mode.

The desired behavior: \`^C\` while waiting should **reserve** the worker instead — transitioning it to the reserved state so the user can use \`/worker:claim\` to pick a specific task without triggering auto-assignment and without a goodbye/hello round-trip.

## Current behavior

- \`^C\` with empty buffer (any state) → submits \`__ctrl_c__\` → \`index.ts\` calls \`workerController.stop()\` → disconnects
- The \`^D\` key (not \`^C\`) is the no-op in the waiting state: CLAUDE.md keyboard rules — "\`^D\` with no visible prompt (state 2, \`ask("")\`) = no-op"

## Desired behavior

\`^C\` while waiting → reserve the worker (send a \`worker_reserve\` message or equivalent), print something like "Reserved — use /worker:claim to pick a task", show a prompt so the user can act.

## Implementation sketch

1. In \`index.ts\`, detect \`__ctrl_c__\` when in the waiting state (worker active, no task) and call a new \`workerController.reserve()\` method instead of \`stop()\`.
2. \`reserve()\` sends a new wire message (e.g. \`worker_reserve\`) or reuses existing \`markReserved\` by adding a reverse of \`worker_ready\`.
3. On the foreman side, handle \`worker_reserve\` by calling \`worker.markReserved()\`.
4. Update CLAUDE.md keyboard rules to document the new behavior.

## Context

Identified during the ready/reserved/assigned state machine refactor (PR #975). The \`worker_ready\` message transitions reserved → ready; this issue tracks the reverse direction (ready → reserved via \`^C\`).

jasoncrawford:
## Summary

- `^C` while a worker is in the **waiting** state (connected, no task, empty prompt) now **reserves** the worker instead of disconnecting
- Sends a new `worker_reserve` wire message to the foreman, which calls `worker.markReserved()` to remove the worker from auto-assignment
- Transitions the worker to a visible `[agent] >` prompt so the user can run `/worker:claim <taskId>`
- A second `^C` from the reserved state calls `stop()` normally

## Changes

- **`shared/wire.ts`**: Added `worker_reserve` to the `WorkerMessage` union type
- **`src/agent/controllers/worker-controller.ts`**: Added `isReserved` getter and `reserve()` method; `_isReserved` resets in `resetSessionState()` and on `task_assigned`
- **`src/agent/index.ts`**: `^C` handler checks `!hasTask() && !isReserved` to route to `reserve()` vs `stop()`; prompt string shows `[agent] >` when reserved
- **`src/foreman/controllers/wss.ts`**: Added `handleWorkerReserve()` and wired it into the message dispatch
- **`CLAUDE.md`**: Updated keyboard rules to document the new `^C`-while-waiting behaviour

## Test plan

- [ ] `reserve()` sends `worker_reserve`, sets `isReserved`, prints message (`worker.test.ts`)
- [ ] `^C` while waiting reserves instead of stops; prompt becomes `[agent] >` afterwards (`worker.mode-switching.test.ts`)
- [ ] Second `^C` (while reserved) stops worker mode (`worker.mode-switching.test.ts`)
- [ ] `handleWorkerReserve` marks worker reserved on foreman side (`foreman.hello-handlers.test.ts`)
- [ ] All 1675 non-DB unit tests pass

Closes #977


shared/wire.ts
```diff
@@ -100,7 +100,8 @@ export type WorkerMessage =
   | { type: "worker_goodbye"; workerId: string; taskId?: string; task_complete?: boolean; stats?: { inputTokens: number; outputTokens: number; costUsd?: number } }
   | { type: "activate_repo"; workerId: string }
   | { type: "claim_task"; workerId: string; taskId: string }
-  | { type: "worker_ready"; workerId: string };
+  | { type: "worker_ready"; workerId: string }
+  | { type: "worker_reserve"; workerId: string };
```
jasoncrawford: let's call this "worker_reserved" to be consistent with the spelling of statuses. or, both worker_ready and worker_reserve could be collapsed into "worker_status", with a status field?

jasoncrawford (commented):
Good call — went with `worker_reserved` to match the status spelling (consistent with `worker_ready` ↔ status `ready`).

Collapsing into `worker_status` would also be clean, but it would require renaming the existing `worker_ready` message that just landed in #980, which feels like a broader scope than this PR. Happy to do that as a follow-up if you prefer.

PR #998, opened 2026-05-03
feat: command aliases

### LINKED ISSUE #994: Command aliases
Commands should be able to have aliases, like "done" or "finished" for "complete".

There can be a list of aliases as part of registering a command.

If registered under a namespace, they all go under it. So when you register "complete" with alias "done", and you're in "worker" namespace, you get "worker:complete" and "worker:done".

Aliases should be mentioned in the command description. After printing the description in autocomplete: "(alias: done)" or "(aliases: done, finished)"

When aliases show up in autocomplete, the description should be the same but it should show what it is an alias for, the canonical name: "(alias for worker:complete)"

Then add the aliases:
* done for worker:complete
* quit for exit

Note that this interacts with #993 to let you type `/done` and have it autocomplete to `/worker:complete`.

jasoncrawford:
## Summary

- `CommandRegistry.register()` accepts an optional `aliases` array; each alias is stored as a separate entry sharing the handler
- Scoped registries apply the namespace prefix to aliases too — `worker.register("complete", { aliases: ["done"] })` creates both `worker:complete` and `worker:done`
- Canonical command descriptions are augmented: `(alias: worker:done)` or `(aliases: name1, name2)`; alias entries say `(alias for worker:complete)`
- Adds `worker:done` (alias for `worker:complete`) and `quit` (alias for `exit`)

## Test plan

- [ ] New `aliases` describe block in `tests/commands.test.ts` covers registration, lookup, scoping, description formatting, and execution
- [ ] New `registered aliases` describe block verifies `worker:done` and `quit` exist with correct `aliasFor` references
- [ ] `tests/repl.autocomplete.test.ts` updated to include the two new alias names in the hardcoded list
- [ ] `tests/helpers.ts` `registerTestCommands()` updated to register aliases, keeping it in sync with production code

Closes #994

### DIFF
--- a/src/agent/controllers/command-controller.ts
+++ b/src/agent/controllers/command-controller.ts
@@ -38,14 +38,14 @@ export type DispatchResult =
 // ── Command filtering ─────────────────────────────────────────────────────────
 
 /**
- * Filter CommandSuggestion objects by substring of name or description.
+ * Filter CommandSuggestion objects by substring of name, alias, or description.
  * Case-insensitive. Empty query returns all commands.
  * Sort order:
- *   1. Prefix matches of any colon-separated suffix, ordered by depth (fewer
- *      leading segments removed = higher priority). `/st` matches `/status`
- *      (depth 0) before `/worker:start` (depth 1) before `/foo:bar:stuff`
- *      (depth 2). `/worker:st` matches `/worker:status` (depth 0) before
- *      `/foo:worker:status` (depth 1).
+ *   1. Prefix matches of any colon-separated suffix of the name or any alias,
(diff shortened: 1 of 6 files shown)


jasoncrawford:
OK, now that I'm trying this out, I feel that it looks dumb to have the aliases appear in autocomplete as separate commands. E.g.:

```
> /quit
  /quit  Exit (alias for exit)
  /exit  Exit (alias: quit)
```

Let's have it so that only the canonical command shows up, like this:

```
> /quit
  /exit  Exit (alias: quit)
```

The autocomplete matching rules should be the same, such that typing `/quit` turns up the exit command as the top match, even if there is `/something:quit` or some other command with quit in a substring.

PR #1001, opened 2026-05-03
feat: inline task ID entry in post-task picker

### LINKED ISSUE #996: Post-task picker: 'Choose a specific task' should accept inline task ID
## Current behavior

After completing a task, the post-task picker shows four options including "Choose a specific task to work on." Selecting it stops the worker and prints a message telling the user to type `/worker:claim <taskId>` manually. This is a dead end — the picker bails out without actually doing anything.

## Desired behavior

The option should accept an inline task ID the way the `AskUserQuestion` "Other:" option works — i.e. using `lastIsTextEntry: true` on the `Picker`. The user selects the option, types a task ID, and the claim flow runs directly from there, without needing to manually invoke a slash command after the picker closes.

## Implementation notes

- `Picker.pick()` supports `lastIsTextEntry: true`; the last option becomes an inline text-entry row
- Rename the option to something like "Claim a specific task..." to signal that input is expected
- On selection, invoke the claim flow directly (same logic as `/worker:claim <taskId>`) rather than calling `stop()` and printing a help message
- Worker should remain in `reserved` state (not `stopped`) while this option is being processed

jasoncrawford:
## Summary

- Renames the post-task "Choose a specific task to work on" option to "Claim a specific task..." and moves it to the last position
- Uses `lastIsTextEntry: true` on the picker so the user can type a task ID inline instead of pressing Enter then manually invoking `/worker:claim`
- When a task ID is entered, the claim flow runs directly (`claim_task` sent to foreman); the worker stays in `reserved` state — no `stop()` call, no `worker_ready`
- Removes the dead-end path that called `stop()` and printed a help message

## Test plan

- [x] New tests: claim option sends `claim_task` with correct ID, worker stays active, returns `task-complete`, no `worker_ready` sent
- [x] Updated option index tests: "stop working" is now option 1, "exit" is now option 2
- [x] Picker label test updated to expect "Claim" instead of "specific task"
- [x] All 250 non-DB tests pass

Closes #996

### DIFF
--- a/src/agent/controllers/worker-controller.ts
+++ b/src/agent/controllers/worker-controller.ts
@@ -9,7 +9,7 @@ import { fmtError } from "../../utils.js";
 import { fmtNum, fmtTaskStats } from "../../../shared/formatters.js";
 import { getConfig } from "../../config.js";
 import type { CommandRegistry } from "./command-controller.js";
-import { Picker } from "../views/picker.js";
+import { Picker, type PickResult } from "../views/picker.js";
 import { WorkspaceController, UserCancelledError } from "./workspace-controller.js";
 
 // ── WorkerDisplay interface ───────────────────────────────────────────────────
@@ -44,9 +44,7 @@ export type WorkerControllerOptions = {
   /** Override the afterTask hook (default: derived from workspaceController.onReset). */
   afterTask?: () => Promise<void>;
   /** Override pick function for repo activation and quit confirmation. */
-  pickFn?: (options: string[]) => Promise<number>;
-  /** Override the post-task completion picker; supports returning a task ID for inline claim. */
-  postTaskPickFn?: (options: string[]) => Promise<number | { type: "text"; text: string }>;
+  pickFn?: (options: string[]) => Promise<number | { type: "text"; text: string }>;
 };
 
 // Messages that must wait for hello_ack before being sent.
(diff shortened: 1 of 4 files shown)


jasoncrawford:
It's weird to have to change the order of options just to allow one of them to have text entry. I didn't want that. Can we enhance the picker so that it supports text entry for any arbitrary option

src/agent/controllers/worker-controller.ts
```diff
@@ -468,28 +470,50 @@ export class WorkerController extends EventEmitter {
...
-      this.sendWorkerReady();
-      this.display.print(c.sageGreen("Waiting for next task..."));
-      return "task-complete";
-    }
-
-    const idx = await pickFn([
+    const options = [
       "Wait to be assigned the next task",
-      "Choose a specific task to work on",
+      "Claim a specific task...",
```
jasoncrawford: this should be "Claim a specific task:" (colon, not elipsis)

src/agent/controllers/worker-controller.ts
```diff
@@ -468,28 +470,50 @@ export class WorkerController extends EventEmitter {
    *
    * Calls sendWorkerReady() only when the user confirms they want to keep waiting —
    * this is what transitions the foreman from "reserved" to auto-assignable.
+   *
+   * The last option ("Claim a specific task...") uses inline text entry so the user
```
jasoncrawford: it's not the last option anymore

src/agent/controllers/worker-controller.ts
```diff
@@ -468,28 +470,50 @@ export class WorkerController extends EventEmitter {
...
+        claimTaskId = r.text.trim();
+      }
+    }
+
+    if (claimTaskId !== undefined) {
+      if (claimTaskId) this.claimAfterTask(claimTaskId);
+      return "task-complete";
+    }
+
+    switch (selectedIndex) {
```
jasoncrawford: This code is incredibly convoluted, with all these fallbacks to a configured pickFn, which only matters in tests, right? and also another fallback if there is no pickFn. We don't need all these fallbacks. Please step back, check your assumptions, and rewrite this in a more straightforward way.

src/agent/controllers/worker-controller.ts
```diff
@@ -468,28 +470,50 @@ export class WorkerController extends EventEmitter {
...
+      if (typeof r === "number") {
+        selectedIndex = r;
+      } else if (r.type === "text") {
+        claimTaskId = r.text.trim();
+      }
+    }
+
+    if (claimTaskId !== undefined) {
+      if (claimTaskId) this.claimAfterTask(claimTaskId);
+      return "task-complete";
```
jasoncrawford: If we claim first, and then return task-complete, aren't we going to be sending a claim message to the foreman before we send the complete message? that seems wrong?

jasoncrawford (commented):
Re comment on lines 512-513: `return "task-complete"` is the TypeScript function return value, not a wire message to the foreman. The `task_complete` wire message is sent much earlier in `completeCurrentTask` (line ~446) before `promptAfterTaskComplete` is ever called, so there's no ordering issue. I'll add a comment to make that clearer, and also address the other three points.

PR #1002, opened 2026-05-03
refactor: collapse worker:start and worker:ready into single command

### LINKED ISSUE #995: Collapse worker:start and :ready
Depends on #994 

The worker:start and ready commands are very similar. Both put you in the same final state. We should just collapse/coalesce them into a single command, `start`, with `ready` as an alias.

* When not in worker mode (not connected to a foreman), connect and go into ready
* When already connected and reserved, go into ready
* When connected with an assigned task, prompt whether to abandon it; if no then cancel, if yes then go into ready

I think this is already the behavior of the ready command. If so we just need to drop 'start', rename 'ready' to 'start', and add 'ready' as an alias.

jasoncrawford:
## Summary

- `/worker:start` now has the full behavior of the old `/worker:ready`: auto-starts if not connected, prompts to abandon an active task, and transitions to the ready (auto-assignable) state
- `/worker:ready` is registered as an alias for `/worker:start`
- The old thin `/worker:start` (which only connected without transitioning to ready) is removed
- Renamed `tests/worker.ready.test.ts` → `tests/worker.start.test.ts` to match

## Test plan

- [ ] Unit tests in `worker.start.test.ts` cover: registration of `worker:start` as canonical and `worker:ready` as alias, auto-start when not active, abandon prompt with active task, happy path (sends `worker_ready`), socket-closed guard
- [ ] `worker.mode-switching.test.ts` updated: second `/worker:start` when already connected transitions to ready (no longer shows "already active")
- [ ] All non-DB tests pass

Closes #995

### DIFF
--- a/src/agent/controllers/worker-controller.ts
+++ b/src/agent/controllers/worker-controller.ts
@@ -583,7 +583,26 @@ export class WorkerController extends EventEmitter {
     });
     registry.register("start", {
       description: "Connect to the foreman and start accepting tasks",
-      handler: async () => { await this.start(); },
+      aliases: ["ready"],
+      handler: async () => {
+        if (!this._isActive) {
+          await this.start();
+          return undefined;
+        }
+        if (this.hasTask()) {
+          const taskInfo = this.getTaskConfirmInfo();
+          if (taskInfo) {
+            this.display.print(c.amber(`\nCurrently assigned task #${taskInfo.taskNumber}. Abandon this task?`));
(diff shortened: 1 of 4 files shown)


src/agent/controllers/worker-controller.ts
```diff
@@ -583,7 +583,26 @@ export class WorkerController extends EventEmitter {
     });
     registry.register("start", {
       description: "Connect to the foreman and start accepting tasks",
```
jasoncrawford: New description: "Start accepting tasks from the foreman"

jasoncrawford: Done.

PR #1013, opened 2026-05-03
refactor: remove pickFn from WorkerControllerOptions; use mock Picker in tests

### LINKED ISSUE #1008: Remove pickFn from WorkerControllerOptions; use mock Picker in tests
From the review in #1005: `WorkerControllerOptions.pickFn` exists only to let tests substitute a mock for picker interactions. The `WorkerController` constructor already accepts a `picker?: Picker` parameter, so tests can pass a mock `Picker` instance instead — no need for a separate escape hatch.

## What to do

- Remove `pickFn` from `WorkerControllerOptions`
- Remove `pickFnOrDefault()` (or inline what it does into the picker path)
- Update tests that currently pass `pickFn` in options to inject a mock `Picker` instance via the constructor instead

## Acceptance criteria

- `WorkerControllerOptions` no longer has a `pickFn` field
- All existing worker tests pass
- No production behaviour changes

jasoncrawford:
## Summary

- Removes `pickFn` from `WorkerControllerOptions` — it existed only as a test escape hatch, duplicating the existing `picker?: Picker` constructor parameter
- Removes `pickFnOrDefault()` helper; each call site now uses `this.picker!.pick(opts)` directly (the four places: `confirmTaskQuit` default, `confirmTaskClaim` default, `/worker:start` abandon prompt, repo-activation prompt)
- Simplifies `postTaskPick`: drops the `pickFn` branch (picker → non-interactive fallback only)
- Adds a `makeMockPicker` helper to the three affected test files; all test setups updated to inject a mock `Picker` instance via the constructor

## Test plan

- [ ] `npm test` — all non-DB unit tests pass
- [ ] No production behaviour changes: the `picker` constructor parameter was already wired up in `index.ts`; only tests are affected

Closes #1008

### DIFF
--- a/src/agent/controllers/worker-controller.ts
+++ b/src/agent/controllers/worker-controller.ts
@@ -361,17 +361,16 @@ export class WorkerController extends EventEmitter {
    */
   async confirmTaskQuit(
     info: TaskConfirmInfo,
-    pickFn: (options: string[]) => Promise<number> = (opts) => this.picker!.pick(opts),
   ): Promise<"quit" | "complete-and-quit" | "cancel"> {
     if (info.issueClosed) {
       this.display.print(c.amber(`\nTask #${info.taskNumber} is closed but not complete. Complete it before exiting?`));
-      const idx = await pickFn(["Yes, complete before exiting", "No, just exit", "Don't exit"]);
+      const idx = await this.picker!.pick(["Yes, complete before exiting", "No, just exit", "Don't exit"]);
       if (idx === 0) return "complete-and-quit";
       if (idx === 1) return "quit";
       return "cancel";
     } else {
       this.display.print(c.amber(`\nTask #${info.taskNumber} is still open. Quitting now will unassign ${info.workerId}. Quit anyway?`));
-      const idx = await pickFn(["No, keep working", "Yes, quit anyway"]);
+      const idx = await this.picker!.pick(["No, keep working", "Yes, quit anyway"]);
       if (idx === 1) return "quit";
       return "cancel";
     }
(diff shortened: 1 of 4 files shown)


src/agent/controllers/worker-controller.ts
```diff
@@ -464,7 +458,7 @@ export class WorkerController extends EventEmitter {
 
   /**
    * After completing a task, prompt the user for what to do next.
-   * When no picker is available (non-interactive / no pickFn injected) defaults to waiting.
+   * When no picker is available (non-interactive mode) defaults to waiting.
```
jasoncrawford: There is no such thing as "non-interactive mode". There should always be a picker available. Can we drop this fallback/default?

src/agent/controllers/worker-controller.ts
```diff
@@ -525,17 +519,10 @@ export class WorkerController extends EventEmitter {
...
       return this.picker.pick(options, { textEntryIndex: 1, textEntryPrefix: "Claim a specific task: " });
     }
-    if (this.options?.pickFn) {
-      const r = await this.options.pickFn(options);
-      return typeof r === "number"
-        ? { type: "selected", index: r }
-        : { type: "other", text: r.text };
-    }
     return { type: "selected", index: 0 }; // non-interactive: default to wait
   }
```
jasoncrawford, line 521: This whole thing should be unnecessary because there should always be a picker

src/agent/controllers/worker-controller.ts
jasoncrawford: Fixed — removed the fallback and the stale comment. The outer session fixture and the handful of `sessionWithAfterTask` instances that go through the ready path now each get a `makeMockPicker(async () => 0)` so they continue to behave as before. The two "with no picker" tests were removed since their coverage is already provided by the "option 0" tests.

src/agent/controllers/worker-controller.ts
jasoncrawford: Done — `postTaskPick` now unconditionally calls `this.picker!.pick(...)`.

src/agent/controllers/worker-controller.ts
```diff
@@ -360,12 +358,10 @@ export class WorkerController extends EventEmitter {
    *
    * Returns 'quit' to proceed without completing, 'complete-and-quit' to mark
    * the task complete then quit, or 'cancel' to stay in the worker.
-   *
-   * An injectable pickFn is accepted so callers can supply a mock in tests.
    */
   async confirmTaskQuit(
     info: TaskConfirmInfo,
-    pickFn: (options: string[]) => Promise<number> = this.pickFnOrDefault(),
+    pickFn: (options: string[]) => Promise<number> = (opts) => this.picker!.pick(opts),
```
jasoncrawford: why does this still have a pickFn argument?

src/agent/controllers/worker-controller.ts
```diff
@@ -389,12 +385,10 @@ export class WorkerController extends EventEmitter {
    *
    * Returns 'complete-and-claim' to complete the current task then claim,
    * 'claim' to abandon the current task and claim, or 'cancel' to do nothing.
-   *
-   * An injectable pickFn is accepted so callers can supply a mock in tests.
    */
   async confirmTaskClaim(
     info: TaskConfirmInfo,
-    pickFn: (options: string[]) => Promise<number> = this.pickFnOrDefault(),
+    pickFn: (options: string[]) => Promise<number> = (opts) => this.picker!.pick(opts),
```
jasoncrawford: why does this still have a pickFn argument?

src/agent/controllers/worker-controller.ts
jasoncrawford: Removed — both methods now call `this.picker!.pick()` directly. The `confirmTaskQuit` tests updated to inject mock Pickers via the constructor.

src/agent/controllers/worker-controller.ts
jasoncrawford: Same — removed.

PR #1019, opened 2026-05-04
feat: resolve slash commands without namespace when unambiguous

### LINKED ISSUE #1017: Allow commands to be invoked without namespace if unambiguous
Right now, if I type `/done` and hit enter, it will invoke `/worker:complete` because that has an alias `/worker:done`, and hitting enter uses the first autocomplete match.

However, if I type `/claim 123` and hit enter, it gives an error: "Unknown command: /claim", even though there is a `/worker:claim` command. That's because when I type space to then enter an argument, the autocomplete goes away. I have to type `/claim<tab>123`, so that when I hit <tab> it is autocompleted to `/worker:claim`.

If an entered slash-command doesn't match anything directly, but it unambiguously matches a command (or alias) without one or more levels of its namespace, we should invoke it.

If it matches more than one command, e.g. if the user enters `/start`, and there's no `/start`, but we have both `/worker:start` and `/workspace:start`, then it should print an error saying "Ambiguous command, which did you mean? /worker:start, /workspace:start"

jasoncrawford:
## Summary

- When a slash command doesn't match directly (e.g. `/claim`), the dispatcher now checks whether any registered command or alias has the entered text as a strict namespace-stripped suffix (e.g. `worker:claim` → `claim`). If exactly one match is found, it is invoked automatically.
- If multiple commands share the same suffix (e.g. `/start` matching both `worker:start` and `workspace:start`), an error is shown: `Ambiguous command: /start — which did you mean? /worker:start, /workspace:start`
- Aliases are included in the suffix search and resolve to their canonical command.

## Test plan

- [x] `parseSlashCommand` tests: unambiguous suffix resolution, alias resolution, ambiguous detection, still-unknown when no match
- [x] `dispatch` tests: args are passed through correctly after suffix resolution, ambiguous command does not fall through to file lookup
- [x] All 98 command-related tests pass; no regressions in the full suite

Closes #1017

### DIFF
--- a/tests/repl.dispatch.test.ts
+++ b/tests/repl.dispatch.test.ts
@@ -1,6 +1,6 @@
 import { describe, it, expect, beforeEach } from "vitest";
+import { CommandRegistry, CommandController } from "../src/agent/controllers/command-controller.js";
 import { registerTestCommands } from "./helpers.js";
-import type { CommandController } from "../src/agent/controllers/command-controller.js";
 
 let registry: CommandController;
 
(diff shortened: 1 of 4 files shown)


tests/repl.dispatch.test.ts
```diff
@@ -111,3 +111,49 @@ describe("dispatchInput", () => {
...
+    reg.scoped("workspace").register("run", { description: "r2", handler: async () => {} });
+    // Even if a file exists for /run, we should get ambiguous_command not query
+    const result = await ctrl.dispatch("/run", (_path) => "file content");
+    expect(result).toEqual({
+      type: "ambiguous_command",
+      command: "run",
+      matches: ["worker:run", "workspace:run"],
+    });
+  });
+});
```
jasoncrawford, line 246: Hmm… I forgot about file lookup. I think that files should be treated the same way as registered commands? If there is a "run.md" or "run/SKILL.md" file in the skills/commands path, that should be matched as if there were a first-class "/run" command registered. Or am I missing some reason why this should not be the behavior? In any case is that a bigger change?

jasoncrawford (commented):
You're right, and it's not a big change. The `dispatch` method already has `readFile` for file lookups, so I'll add `listDir` as an optional third param and do a second suffix-matching pass over the file-based commands returned by `listCommandNames`. Registry-only calls to `parseSlashCommand` stay unchanged; the file check happens in `dispatch` after the registry lookup comes back empty. I'll push it now.

jasoncrawford (commented):
This looks to be right, but is there a reason it has to be so complicated? Is it because we're walking the directory at command invocation time? Otherwise we could just find all the commands on startup and register them, and then have everything in a single registry.

jasoncrawford (commented):
The complexity comes down to one thing: file-based commands can't currently fit in the registry because they don't have side-effect handlers — they produce `{ type: "query", prompt }` that `index.ts` sends to `runPrompt`, whereas registry commands produce `{ type: "command" }` that call `registry.execute`. To unify them we'd need one of:

1. **New `HandlerResult` variant** — add `{ type: "query"; prompt: string }` to `HandlerResult`, then `index.ts` handles it alongside the existing `"exit"` / `"task-complete"` returns. File commands get registered with handlers that load their content and return `{ type: "query", prompt }`. `dispatch` stays simple and registry-only.

2. **Late registration** — register file commands after `runPrompt` is available in `BrunelAgent.start()`, with handlers that close over it.

Option 1 is the cleaner change. Happy to pursue it (either now or as a follow-up). The current approach does re-walk the directory on each dispatch but it's the same work `suggest()` already does for autocomplete. What would you prefer?

PR #1051, opened 2026-05-06
Worker: source GitHub token from gh CLI/env/prompt; use extraHeader for git auth

### LINKED ISSUE #1042: Worker: source GitHub token from gh CLI / env var / prompt; use extraHeader for git auth

## Token sourcing

Workers obtain their GitHub token in this order (additive — existing env var path still works):

1. `gh auth token` — zero config if GitHub CLI is installed and authenticated
2. `GITHUB_TOKEN` or `GH_TOKEN` environment variable (existing behaviour)
3. Prompt the user with a link to create a fine-grained PAT (`contents: read/write`, `pull_requests: read/write`, `issues: read/write`, `metadata: read`)

The resolved token is sent in `worker_hello.githubToken` and held for use in git operations.

## Git auth: extraHeader instead of token-in-URL

Replace the current approach of embedding the token in the git remote URL with `http.extraHeader`:

```bash
git config --local http.https://github.com/.extraheader "Authorization: Bearer {token}"
```

The remote URL stays clean (`https://github.com/owner/repo.git`). Token never appears in `git remote -v` or process listings.

## Notes

- Can start immediately — no dependencies on other Phase 2 issues
- Existing workers that already have a token via env var or config continue to work; this just adds `gh auth token` as a higher-priority source

jasoncrawford:
## Summary

- **Token sourcing**: Workers now resolve their GitHub token in priority order: `gh auth token` CLI → `GITHUB_TOKEN`/`GH_TOKEN` env vars → interactive user prompt (TTY only). Existing env var behavior is preserved; `gh auth token` is added as a zero-config highest-priority source.
- **Secure git auth**: Replaces token-in-URL approach (`https://{token}@github.com/...`) with `git config --local http.https://github.com/.extraheader "Authorization: Bearer {token}"` set after clone. The remote URL stays clean — token never appears in `git remote -v` or process listings.
- **Wire protocol**: `worker_hello` now includes an optional `githubToken` field so the foreman knows which token the worker resolved.
- **Config**: `githubToken` is now optional in the schema (workers resolve it dynamically); the foreman validates it at startup and exits with a clear error if absent.

## Test plan

- [ ] `npm test` passes (all non-DB tests green)
- [ ] `npx tsc --noEmit` passes (no type errors)
- [ ] `npm run lint` passes (no new errors)
- [ ] New `tests/github-token.test.ts` covers `resolveGithubTokenFromCli` and `resolveGithubToken`
- [ ] `tests/workspace.test.ts` covers `http.extraHeader` being set after clone, and that clone URL is clean
- [ ] `tests/worker.test.ts` covers `githubToken` appearing in `worker_hello` when provided
- [ ] Manual: start a worker with no `GITHUB_TOKEN` set but `gh` authenticated → worker starts cleanly
- [ ] Manual: start a worker with no `GITHUB_TOKEN` and no `gh` auth → user is prompted for a PAT

Closes #1042


src/agent/models/github-token.ts
jasoncrawford, line 1: I see bad patterns here. Functions given as arguments, (presumably to mock for tests), special hashes of options, exported loose functions instead of an object… please clean all this up. I suggest exporting a first-class object, and having tests replace/mock its methods if needed.

src/agent/index.ts
```diff
@@ -67,17 +69,44 @@ export class BrunelAgent {
...
+    return new Promise((resolve) => {
+      const rl = readline.createInterface({ input: process.stdin, output: process.stdout, terminal: false });
+      process.stdout.write("Enter GitHub personal access token: ");
+      rl.once("line", (line) => {
+        rl.close();
+        resolve(line.trim() || null);
+      });
+      rl.once("close", () => resolve(null));
+    });
+  }
```
jasoncrawford: What's the UX here… does the user have to respond to this prompt every time? If so isn't it better to instruct them to put the token in the environment, and/or to install the `gh` cli, so they don't have to do this?

src/agent/index.ts
```diff
@@ -67,17 +69,44 @@ export class BrunelAgent {
...
+    // git config http.extraHeader (keeping auth out of the remote URL).
     const { config } = this;
-    const repoUrl = config.repoUrl ?? (repo && config.githubToken
-      ? `https://${config.githubToken}@github.com/${repo}.git`
+    const ghCliToken = await resolveGithubTokenFromCli();
+    const githubToken = await resolveGithubToken({
+      cliToken: ghCliToken,
+      configToken: config.githubToken,
+      promptFn: process.stdin.isTTY ? () => this._promptForGithubToken() : undefined,
+    });
```
jasoncrawford: this logic should be encapsulated inside the github-token module

src/agent/index.ts
```diff
@@ -86,7 +115,7 @@ export class BrunelAgent {
         }
       : undefined;
     const workspace = workspaceCfg
-      ? new Workspace(workspaceCfg.workspaceDir, this.agentStatus.agentId, workspaceCfg.repoUrl, this.originalCwd, this.confirm)
+      ? new Workspace(workspaceCfg.workspaceDir, this.agentStatus.agentId, workspaceCfg.repoUrl, this.originalCwd, this.confirm, undefined, undefined, githubToken ?? undefined)
```
jasoncrawford: This is a big ugly argument list… if nothing else we should move githubToken in front of the exec and npm arguments, so that we can leave those out. Better would be to get rid of the exec and npm arguments and mock those in tests, but we can do that in a separate PR later.

src/foreman/index.ts
```diff
@@ -19,6 +19,11 @@ const isMain = process.argv[1] === fileURLToPath(import.meta.url);
 if (isMain) {
   const config = await loadConfig(process.argv);
 
+  if (!config.githubToken) {
+    log("ERROR GitHub token required. Set GITHUB_TOKEN or BRUNEL_GITHUB_TOKEN.");
+    process.exit(1);
+  }
```
jasoncrawford, line 25: Hmm… aren't we going to start using the GH app for auth, and not require this on the server?

CLAUDE.md
```diff
@@ -28,7 +28,7 @@ A unified REPL + worker loop. `index.ts` is the composition root; `AgentControll
 
 Follows MVC with three subdirectories:
 
-- **Models** (`models/`) — `Workspace` (git/npm workspace management), `Settings` (runtime-settable preferences: model, effort, permissions), `QueryStats` (token usage/turn counts and API cost from SDK messages), `AgentStatus` (pure state model for worker status — emits `"change"` on updates, subscribed to by `Display` for reactive redraws; also owns static git/id utilities `getCurrentBranch`, `getRemoteRepo`, `generateAgentId`)
+- **Models** (`models/`) — `Workspace` (git/npm workspace management; uses `http.extraHeader` for git auth rather than embedding tokens in clone URLs), `github-token.ts` (GitHub token resolution: `gh auth token` CLI → `GITHUB_TOKEN`/`GH_TOKEN` env vars → interactive prompt; resolved token passed to `Workspace` and sent in `worker_hello`), `Settings` (runtime-settable preferences: model, effort, permissions), `QueryStats` (token usage/turn counts and API cost from SDK messages), `AgentStatus` (pure state model for worker status — emits `"change"` on updates, subscribed to by `Display` for reactive redraws; also owns static git/id utilities `getCurrentBranch`, `getRemoteRepo`, `generateAgentId`)
```
jasoncrawford: feels like a bit too much detail; keep implementation details out of CLAUDE.md

src/agent/models/github-token.ts
jasoncrawford: Done — replaced with a `GithubToken` object exporting `fromCli()` and `resolve()`. Tests use `vi.spyOn(GithubToken, 'fromCli')` and `vi.mock(..., () => ({ ...actual, GithubToken: { ...actual.GithubToken, fromCli: vi.fn() } }))` instead of constructor injection.

src/agent/index.ts
jasoncrawford: Removed the interactive prompt entirely. Now if no token is found we print a one-line hint: "Run `gh auth login` or set GITHUB_TOKEN to enable workspace cloning." and continue (the worker still works without a workspace).

src/agent/index.ts
jasoncrawford: Done — `index.ts` now just calls `await GithubToken.resolve(config.githubToken)`. The full CLI → env-var chain lives inside the module.

src/agent/index.ts
jasoncrawford: Done — `githubToken` is now position 6 (before `exec` and `npm`). The call site is now `new Workspace(..., this.confirm, githubToken ?? undefined)` with no trailing `undefined, undefined`.

src/foreman/index.ts
jasoncrawford: Good question — I've left the check in place for now since the foreman actively uses `config.githubToken` today. Happy to remove it once the GH App auth lands; want me to open a follow-up issue for that?

CLAUDE.md
jasoncrawford: Trimmed — the `http.extraHeader` detail is gone, and the entry now just says "GitHub token resolution: gh CLI → env vars".

src/config.ts
```diff
@@ -27,8 +27,8 @@ const BrunelConfigSchema = z.object({
 
   /** GitHub repo in "owner/repo" format. Optional; workers detect it from git remote automatically. */
   githubRepo:     z.string().min(1).optional(),
-  /** GitHub personal access token with `repo` scope. Required. Prefer env var over config file. */
-  githubToken:    z.string().min(1),
+  /** GitHub personal access token with `repo` scope. Prefer env var over config file. Workers can also resolve this via `gh auth token` or an interactive prompt at startup. */
```
jasoncrawford: interactive prompt is now gone

src/agent/models/github-token.ts
```diff
@@ -0,0 +1,30 @@
...
+   * Resolve a GitHub token in priority order:
+   * 1. gh CLI (`gh auth token`)
+   * 2. configToken (from GITHUB_TOKEN / BRUNEL_GITHUB_TOKEN env / config file)
+   *
+   * Returns null if no token is available.
+   */
+  async resolve(configToken?: string): Promise<string | null> {
+    const cliToken = await this.fromCli();
+    if (cliToken) return cliToken;
+    if (configToken) return configToken;
```
jasoncrawford: config should override `gh auth token`, check it first

src/agent/index.ts
```diff
@@ -67,17 +68,26 @@ export class BrunelAgent {
...
     // Show current branch in the minimal status bar before worker mode activates.
     await this.agentStatus.refreshBranch();
 
-    // Build workspace config after repo detection. repoUrl can be set explicitly
-    // in config; otherwise it's derived from the detected git remote + token.
+    // Resolve GitHub token: gh CLI → env/config.
     const { config } = this;
-    const repoUrl = config.repoUrl ?? (repo && config.githubToken
-      ? `https://${config.githubToken}@github.com/${repo}.git`
+    const githubToken = await GithubToken.resolve(config.githubToken);
```
jasoncrawford: It ls a little odd to extract the token from the config, pass it in to resolve, and then have it just return the argument. It seems like either:

* index.ts should own the fallback logic, in which case it checks config, and then uses the model object to run `gh auth token` if necessary. At that point GithubToken is pretty anemic… the one little function could just live in Workspace, and all this logic could move in there?
* Or: we create the model here with `token = new GithubToken(config)`, and then call `token.resolve()` which does the full logic.

src/agent/models/github-token.ts
```diff
@@ -0,0 +1,30 @@
+import { execFile as execFileCb } from "node:child_process";
+import { promisify } from "node:util";
+
+const execFileAsync = promisify(execFileCb);
+
+export const GithubToken = {
```
jasoncrawford: this isn't really an object, all the methods are static

CLAUDE.md
```diff
@@ -28,7 +28,7 @@ A unified REPL + worker loop. `index.ts` is the composition root; `AgentControll
 
 Follows MVC with three subdirectories:
 
-- **Models** (`models/`) — `Workspace` (git/npm workspace management), `Settings` (runtime-settable preferences: model, effort, permissions), `QueryStats` (token usage/turn counts and API cost from SDK messages), `AgentStatus` (pure state model for worker status — emits `"change"` on updates, subscribed to by `Display` for reactive redraws; also owns static git/id utilities `getCurrentBranch`, `getRemoteRepo`, `generateAgentId`)
+- **Models** (`models/`) — `Workspace` (git/npm workspace management), `github-token.ts` (GitHub token resolution: `gh auth token` CLI → `GITHUB_TOKEN`/`GH_TOKEN` env vars; resolved token passed to `Workspace` and sent in `worker_hello`), `Settings` (runtime-settable preferences: model, effort, permissions), `QueryStats` (token usage/turn counts and API cost from SDK messages), `AgentStatus` (pure state model for worker status — emits `"change"` on updates, subscribed to by `Display` for reactive redraws; also owns static git/id utilities `getCurrentBranch`, `getRemoteRepo`, `generateAgentId`)
```
jasoncrawford: no need to include token resolution details here; they'll just get out of sync with the code

src/config.ts
jasoncrawford: Updated the JSDoc to say 'Workers also fall back to `gh auth token` if this is unset.'

src/agent/models/github-token.ts
jasoncrawford: Done — config/env token is now checked first; gh CLI is the fallback.

src/agent/index.ts
jasoncrawford: Went with Option B: `new GithubToken(config.githubToken).resolve()`. The config token is injected at construction so `resolve()` owns the full logic without needing arguments.

src/agent/models/github-token.ts
jasoncrawford: Refactored to a proper class: `new GithubToken(configToken?)` with a `resolve()` instance method.

CLAUDE.md
jasoncrawford: Removed `github-token.ts` from the CLAUDE.md models list entirely.

src/agent/models/github-token.ts
```diff
@@ -0,0 +1,29 @@
+import { execFile as execFileCb } from "node:child_process";
+import { promisify } from "node:util";
+
+const execFileAsync = promisify(execFileCb);
+
+export class GithubToken {
+  constructor(private readonly configToken?: string) {}
```
jasoncrawford: I think it should take the config object and know how to extract the right config option.

jasoncrawford: Done — constructor now takes `{ githubToken?: string }` and `resolve()` reads it directly. Call site is `new GithubToken(config).resolve()`.

PR #1053, opened 2026-05-08
Add App credentials to config; GithubClient installation-token minting

### LINKED ISSUE #1039: Foreman: App credentials in config; GithubClient installation-token minting

## Config changes

Add three optional fields to foreman config:
- `appId: string` — GitHub App ID
- `appPrivateKey: string` — PEM private key for signing JWTs
- `appWebhookSecret: string` — for verifying incoming webhook signatures

All optional. When absent, existing behaviour is unchanged.

## GithubClient changes

Add `mintInstallationToken(installationGithubId: number): Promise<string>` — signs a JWT with the App private key, exchanges it for an installation access token via `POST /app/installations/{id}/access_tokens`. Returns the token string.

All existing `GithubClient` methods continue to accept a token parameter (or read from config). When the App is configured, callers pass an installation token; when not, they fall back to the personal `githubToken` from config.

## Notes

- No behaviour changes for existing self-hosted users
- Additive: App config fields are optional; `githubToken` in config remains valid

Depends on #1038

jasoncrawford:
## Summary

- Adds three optional foreman config fields: `appId`, `appPrivateKey`, `appWebhookSecret` — all optional, no behaviour change for existing users
- `appPrivateKey` and `appWebhookSecret` are added to `SECRET_KEYS` (warning logged if found in a config file)
- Adds `GithubClient.mintInstallationToken(installationGithubId)` — mints a short-lived RS256 JWT from the App credentials, exchanges it for a GitHub installation access token via `POST /app/installations/{id}/access_tokens`, and returns the token string
- Existing `GithubClient` methods (`fetchIssues`, `fetchIssueStates`, `fetchNativeBlockers`) gain an optional `token` parameter; callers can pass an installation token, or the method falls back to `githubToken` from config

## Test plan

- [ ] Config tests: `appId`, `appPrivateKey`, `appWebhookSecret` all default to undefined; resolve from `BRUNEL_APP_*` env vars and `--app-*` CLI flags; secrets warn when found in file config
- [ ] `mintInstallationToken` tests: calls correct endpoint, JWT has correct `iss`/`iat`/`exp` claims, returns the token string, throws on non-ok response and when credentials missing
- [ ] Optional token parameter tests for existing methods
- [ ] All 1835 non-DB unit tests pass; type check and lint clean

Closes #1039

### DIFF
--- a/src/foreman/clients/github.ts
+++ b/src/foreman/clients/github.ts
@@ -1,3 +1,4 @@
+import { createSign } from "node:crypto";
 import { getConfig } from "../../config.js";
 
 // ── GitHub API helpers ────────────────────────────────────────────────────────
@@ -8,6 +9,20 @@ function ghHeaders(token: string) {
     Accept: "application/vnd.github+json",
     "X-GitHub-Api-Version": "2022-11-28",
   };
+}
+
+function mintAppJwt(appId: string, privateKey: string): string {
+  const now = Math.floor(Date.now() / 1000);
+  const header = Buffer.from(JSON.stringify({ alg: "RS256", typ: "JWT" })).toString("base64url");
+  const payload = Buffer.from(JSON.stringify({
+    iss: appId,
+    iat: now - 60,
+    exp: now + 540,
+  })).toString("base64url");
+  const data = `${header}.${payload}`;
+  const sign = createSign("RSA-SHA256");
+  sign.update(data);
+  return `${data}.${sign.sign(privateKey, "base64url")}`;
 }
 
 // ── Types ─────────────────────────────────────────────────────────────────────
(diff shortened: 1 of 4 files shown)


src/foreman/clients/github.ts
jasoncrawford, line 1: I don't like the design of passing in an optional token argument on every method. Why don't we pass an optional installationGithubId on the constructor? Then the mintInstallationToken can use that id, and the other methods can use whatever token it generates.

jasoncrawford (commented):
This PR feels incomplete / not self-contained. We throw in some code to mint installation tokens, but we don't actually use it anywhere?

src/foreman/clients/github.ts
jasoncrawford: Done — moved `installationGithubId` to the constructor. A private `resolveToken()` auto-mints when an ID is set and falls back to config `githubToken` otherwise. `mintInstallationToken()` now takes no arguments.

jasoncrawford:
On the "not used anywhere" concern: this issue is deliberately foundational — the next issue (webhook handling for `installation`/`installation_repositories` events) will be the first caller, constructing clients as `new GithubClient(repo.fullName, installation.githubId)`. With the constructor redesign, wiring it up at that point is a one-line change in `TaskManager`.

PR #1058, opened 2026-05-08
Foreman: worker auth via GitHub token

### LINKED ISSUE #1041: Foreman: worker auth via GitHub token (additive)

## Changes

`worker_hello` gains an optional `githubToken` field. When the App is configured and the worker's repo has an `installation_id`, the foreman verifies push access:

```
GET /repos/{owner}/{repo}/collaborators/{username}/permission
```

Called with an installation token (minted for the repo). If the response is not `push` or `admin`, the foreman sends a fatal `foreman_error` and closes the connection.

To get the username, call `GET /user` with the worker's personal token before the permission check.

## Existing path unchanged

When `githubToken` is absent in `worker_hello`, or when the App is not configured, the foreman falls back to the existing `workerSecret` verification. Both paths remain valid during this milestone.

Depends on #1039

jasoncrawford:
## Summary

- When the GitHub App is configured (`appId` + `appPrivateKey`) and the worker's repo has an `installation_id`, the foreman verifies push access on `worker_hello`: calls `GET /user` with the worker's personal token to get their username, then `GET /repos/{owner}/{repo}/collaborators/{username}/permission` with a minted installation token
- Workers without `push` or `admin` permission receive a fatal `foreman_error`
- Falls back to existing `workerSecret` verification when `githubToken` is absent, the App is not configured, or the repo has no `installation_id`
- Adds `GithubClient.fetchUserLogin()` and `GithubClient.verifyPushAccess()` to the GitHub client
- Extends the in-memory test DB to support `installations` table and `installation_id` on repos; adds `seedRepoWithInstallation()` test helper

## Test plan

- [ ] New tests in `tests/foreman.github.test.ts` cover `fetchUserLogin` and `verifyPushAccess` (success, failure, wrong permission)
- [ ] New tests in `tests/foreman.websocket.test.ts` cover: accept with push/admin, reject with insufficient access, fall back to workerSecret when App not configured / no installation / no githubToken, fatal error on API failure
- [ ] All 1864 existing tests continue to pass
- [ ] TypeScript clean (`npx tsc --noEmit`)

Closes #1041


src/foreman/controllers/wss.ts
```diff
@@ -354,6 +351,34 @@ export class ForemanWss {
...
+      if (!authorized) {
+        log(`[worker ${shortWorkerId(workerId)}] GitHub token auth rejected — insufficient repo access`);
+        this.sendError(ws, "Insufficient repository access — push or admin permission required", true, workerId, repo.id);
+        return;
+      }
+    } else {
+      if (this.config.workerSecret && msg.workerSecret !== this.config.workerSecret) {
+        ws.close(4001, "unauthorized");
+        return;
+      }
```
jasoncrawford, line 380: why in one case do we send an error message, and in another case we close the connection? what's the right thing to do here?

CLAUDE.md
```diff
@@ -117,6 +121,6 @@ See **`docs/type-system.md`** for the full design. In brief: one server model cl
...
+- **Browser tests**: the server is shared across all tests — use unique issue numbers per test rather than resetting server state. Test helper endpoints (`/test/*`) must await any async side effects (e.g. WebSocket close handshakes, DB writes) before sending their HTTP response, so that callers can rely on the side effects being complete when the `await fetch(...)` resolves. The in-memory DB shim (`tests/helpers/memory-db.ts`) covers `tasks`, `repos`, `workers`, `webhook_events`, and `installations`; `foreman_messages` is stubbed as empty. For features that read `foreman_messages`, use `page.route("**/api/...**", handler)` to mock the API response rather than relying on real DB rows. Scroll-triggered UI (IntersectionObserver) may auto-fire on load in compact CI viewports — use `page.evaluate(() => window.scrollTo(0, document.body.scrollHeight))` to reliably trigger it rather than waiting for a button to appear.
```
jasoncrawford: Let's reduce the amount of details here so we don't have to keep this up to date with every code change. Just make this a high-level description.

jasoncrawford (commented):
Good catches, both addressed:

**CLAUDE.md test section** — stripped it back to high-level guidance (patterns and gotchas only), no specific function names or table lists.

**Auth path inconsistency** — `workerSecret` failure now also uses `sendError(fatal: true)` instead of `ws.close(4001)`. That makes all fatal rejections in `handleWorkerHello` consistent: they all send a `foreman_error` message so the human at the terminal sees a useful reason, and the worker closes itself in response. The workerSecret test is updated to assert on the `foreman_error` message rather than the close event.

PR #1059, opened 2026-05-08
Foreman: handle installation webhooks; auto-activate direct-repo installs

### LINKED ISSUE #1040: Foreman: handle installation webhooks; auto-activate direct-repo installs

## Webhook events to handle

- `installation` with `action: created` — create `Installation` record; if `target_type` is `User` (direct repo install), also link and activate all repos in `repositories` array, seeding tasks from open `brunel:ready` issues using an installation token
- `installation` with `action: deleted` — delete `Installation` record; set `installation_id = null` and deactivate all linked repos (preserve existing task records)
- `installation_repositories` with `action: added` — for direct-repo installs, link and activate the new repos; for org installs, no-op (activation deferred to first worker connect)
- `installation_repositories` with `action: removed` — unlink and deactivate the removed repos

## Org install / lazy activation

When a worker connects claiming a repo under an org that has an `Installation` record, the foreman:
1. Looks up the installation by `account_login`
2. Sets `repos.installation_id`
3. Activates the repo and seeds tasks

This happens in the existing `worker_hello` / `hello_ack` flow, not here — noted for context.

## Notes

- Self-hosted users receive no `installation` webhooks; this code path is never triggered for them
- Seeding uses installation token; falls back to personal `githubToken` if App not configured (for self-hosted)

Depends on #1039

jasoncrawford:
## Summary

- Adds `InstallationsController` to handle GitHub App installation lifecycle webhooks (`installation.created/deleted`, `installation_repositories.added/removed`)
- `installation.created` for User (direct-repo) installs: creates `Installation` record, then links, activates, and seeds tasks for each repo using an installation token
- `installation.deleted`: deletes the `Installation` record and unlinks/deactivates all associated repos (preserving task records)
- `installation_repositories.added` for User installs: link/activate/seed new repos; org installs are no-op (lazy activation deferred to worker connect)
- `installation_repositories.removed`: unlinks and deactivates removed repos
- `ForemanWss.handleActivateRepo()` now passes the installation GitHub ID to `loadIssuesFromGithub()` so worker-triggered activation also uses the installation token
- In-memory DB shim extended to cover the `installations` table (fixes a latent bug where `installation_id` was missing from new repo rows)

## Test plan

- [ ] 16 new unit tests in `tests/foreman.installations.test.ts` covering all four handler paths (created User/Org, deleted, repos added User/Org, repos removed) — all pass
- [ ] Full test suite: 1865 tests pass, 0 new failures (5 pre-existing `db.*.test.ts` failures require live Supabase)
- [ ] `npx tsc --noEmit` exits 0

Closes #1040


src/foreman/models/task-manager.ts
```diff
@@ -437,10 +437,14 @@ export class TaskManager extends EventEmitter {
   }
 
   /** Fetch brunel:ready issues from GitHub and load deps.
-   *  Called at startup after loadActiveTasksFromDb. */
-  async loadIssuesFromGithub(): Promise<void> {
+   *  Called at startup after loadActiveTasksFromDb.
+   *  Pass installationGithubId to mint an installation token instead of the personal token. */
+  async loadIssuesFromGithub(installationGithubId?: number): Promise<void> {
```
jasoncrawford: why is this an optional arg? shouldn't we get it off the repo object?

jasoncrawford: and we should be able to do that at the time the TaskManager makes its client, so we don't have to make one inline

CLAUDE.md
jasoncrawford, line 1: too much implementation detail here, make this a high-level guide not a list of symbols that we have to keep up to date and in sync with the code

src/foreman/models/repo.ts
```diff
@@ -78,6 +78,33 @@ export class Repo extends ActiveRecord {
...
+
+  /** Sets installation_id to the given value and persists. Returns the updated Repo instance. */
+  async linkInstallation(installationId: number): Promise<Repo> {
+    return this.update({ installation_id: installationId });
+  }
+
+  /** Clears installation_id and persists. Returns the updated Repo instance. */
+  async unlink(): Promise<Repo> {
+    return this.update({ installation_id: null });
+  }
```
jasoncrawford, line 100: Why does deactivate update the local `status` property but the link/unlink methods don't update the local `installationId`?

Also `linkInstallation` is not parallel with `unlink`, call it `unlinkInstallation` for clarity

jasoncrawford (commented):
When the webhook messages are received, will they be recorded in the webhook database table? I want to make sure we have a log of what happened.

jasoncrawford (commented):
**Webhook logging**: Yes, installation events are fully logged. `routeEvent()` calls `WebhookEvent.log()` unconditionally at step 2 (after routing) — installation events flow through the same path as all others, just without a linked task or repo ID.

**Code changes** (just pushed):

- `Repo.unlink()` → `Repo.unlinkInstallation()` for symmetry with `linkInstallation()`
- `Repo.installationId` is now mutable; `linkInstallation`/`unlinkInstallation` update it in-memory, consistent with how `activate`/`deactivate` handle `status`
- Removed `installationGithubId?` param from `loadIssuesFromGithub` — the method now resolves `await this.repo.installation` internally. All callers simplified to `loadIssuesFromGithub()` with no args
- CLAUDE.md architecture section trimmed to high-level descriptions without symbol enumerations

jasoncrawford (commented):
Addressed all review points — see commit 5e5d1cd. Summary:

- **Webhook logging**: yes, installation events are logged by the existing unconditional `WebhookEvent.log()` call in `routeEvent()`.
- `Repo.unlink()` renamed to `unlinkInstallation()` for symmetry
- `Repo.installationId` made mutable; `linkInstallation`/`unlinkInstallation` now update the in-memory property (matching how `activate`/`deactivate` handle `status`)
- Removed `installationGithubId?` param from `loadIssuesFromGithub` — method now resolves `await this.repo.installation` internally; all callers simplified to no-arg call
- CLAUDE.md trimmed to high-level descriptions

jasoncrawford:
Addressed all review points — see commit 5e5d1cd:

- **Webhook logging**: yes, installation events are logged by the existing unconditional `WebhookEvent.log()` call in `routeEvent()` — same path as all other webhooks, just without a linked task/repo.
- `Repo.unlink()` → `Repo.unlinkInstallation()` for symmetry
- `Repo.installationId` made mutable; `linkInstallation`/`unlinkInstallation` now update the in-memory property (matching how `activate`/`deactivate` handle `status`)
- Removed `installationGithubId?` param from `loadIssuesFromGithub` — method now resolves `await this.repo.installation` internally; all callers simplified to no-arg call
- CLAUDE.md trimmed to high-level descriptions

src/foreman/models/task-manager.ts
```diff
@@ -437,10 +437,12 @@ export class TaskManager extends EventEmitter {
   }
 
   /** Fetch brunel:ready issues from GitHub and load deps.
-   *  Called at startup after loadActiveTasksFromDb. */
+   *  Called at startup after loadActiveTasksFromDb, and on installation/activation events. */
   async loadIssuesFromGithub(): Promise<void> {
     const repo = this.repo.fullName;
-    const issues = await this.github.fetchIssues();
+    const installation = await this.repo.installation;
+    const github = installation ? new GithubClient(repo, installation.githubId) : this.github;
```
jasoncrawford: Why are we making a special client here, and only here? I think we should do this in the constructor, when we make the GithubClient, just make the correct one based on the repo and whether it has an installation or not. Then we don't need this logic here, we can just use this.github.

jasoncrawford:
Done — see commit 510ed74. Made `github` a private getter on `TaskManager` that picks the right client from `this.repo.installationId` (sync, no async fetch needed since `installationId` is already a mutable field updated by `linkInstallation`). The personal-token client is cached in `_github`; installation clients are created on access. `loadIssuesFromGithub` now just does `const github = this.github`.

src/foreman/models/task-manager.ts
```diff
@@ -437,10 +439,11 @@ export class TaskManager extends EventEmitter {
   }
 
   /** Fetch brunel:ready issues from GitHub and load deps.
-   *  Called at startup after loadActiveTasksFromDb. */
+   *  Called at startup after loadActiveTasksFromDb, and on installation/activation events. */
   async loadIssuesFromGithub(): Promise<void> {
     const repo = this.repo.fullName;
-    const issues = await this.github.fetchIssues();
+    const github = this.github;
+    const issues = await github.fetchIssues();
```
jasoncrawford: this change is needless now

src/foreman/models/task-manager.ts
```diff
@@ -464,7 +467,7 @@ export class TaskManager extends EventEmitter {
     }
 
     if (allBlockerNumbers.size > 0) {
-      const states = await this.github.fetchIssueStates(Array.from(allBlockerNumbers));
+      const states = await github.fetchIssueStates(Array.from(allBlockerNumbers));
```
jasoncrawford: this change is needless now

PR #1083, opened 2026-05-09
refactor: cleaner routing for server requests

### LINKED ISSUE #1064: Cleaner routing for server requests
## Problem

`ForemanWss` (~810 lines) conflates three concerns: **transport** (WebSocket connection lifecycle, heartbeats), **routing** (dispatching events/messages to handlers), and **business logic** (the handler implementations). Webhook and worker-message dispatch are ad-hoc if/else chains. All API handlers live inline in the `HttpServer` constructor.

## Goal

Separate these three concerns cleanly:

```
WebhookRouter   → @octokit/webhooks registry  →  WebhookController.handleX()  →  RouteResult
WorkerMsgRouter → Map<type, handler>           →  WorkerController.handleX(worker, msg, messenger)
HttpRouter      → Hono                         →  ApiController
```

### 1. ApiController

Extract all `/api/*` route handlers out of `HttpServer` into an `ApiController`. `HttpServer` stays as pure routing infrastructure; it calls `apiController.register(app)` to mount routes.

### 2. WebhookController

Extract `routeEvent`, `routePrEvent`, `routePrReviewEvent`, `routeCheckEvent`, `routeIssueEvent`, and `applyIssueEffects` from `ForemanWss` into a `WebhookController`.

Register handlers via `webhooks.on()`:
- `"issues"` → issue events (label/unlabel, close, reopen, edit, comment)
- `"pull_request"` → PR opened/closed/edited/synchronize
- `"pull_request_review"` / `"pull_request_review_comment"` → review events
- `"check_run"` / `"check_suite"` → CI check events

**Dev mode unification**: currently dev mode (no webhook secret) bypasses `@octokit/webhooks` entirely via a manual `routeEvent` call. Replace this with `webhooks.receive({ id, name, payload })`, which dispatches without verifying — one registry, two entry points (verified vs. unverified), no parallel dispatch logic.

Handlers return a `RouteResult`. The routing layer owns the shared post-handler steps: logging to DB (`WebhookEvent.log`), forwarding to worker (`forwardEvent`), and admin dashboard broadcasting.

### 3. WorkerController + WorkerMessenger

Extract `handleWorkerHello`, `handleTaskComplete`, `handleWorkerGoodbye`, `handleActivateRepo`, `handleClaimTask`, `handleWorkerReady`, `handleWorkerReserve` from `ForemanWss` into a `WorkerController`.

Introduce a `WorkerMessenger` service (analogous to a renderer in MVC) injected into the controller. Instead of calling `this.sendMsg(...)` directly, handlers call `messenger.send(worker, msg)` and `messenger.sendError(...)`. `WorkerMessenger` wraps the WebSocket send infrastructure and handles logging/broadcasting sent messages.

### 4. WorkerMsgRouter

Replace the if/else cascade in `ws.on("message")` with a `Map<Wire.WorkerMessage["type"], handler>` registry. `WorkerController` registers its handlers at construction time. The router:
1. Parses and logs the incoming message
2. Looks up and calls the handler
3. Calls `assignWork()` as a post-step (cross-cutting concern — belongs in the router, not individual handlers)

### 5. ForemanWss (thin transport layer)

After extraction, `ForemanWss` owns only: WebSocket server setup, ping/pong heartbeat, upgrade handler, connect/disconnect lifecycle, and the `WorkerMessenger` send infrastructure. All business logic delegated to `WorkerController` via the router.

## Out of scope

- Admin WebSocket (`AdminWss` is broadcast-only; no routing needed)
- Any wire protocol changes
- Any behavior changes — pure refactoring

jasoncrawford:
## Summary

- Extract `WorkerMessenger` from `ForemanWss`: owns all outbound send infrastructure (analogous to a renderer in MVC), including admin broadcast log events
- Extract `ApiController` from `HttpServer`: all `/api/*` REST route handlers mounted via `ApiController.register(app)`
- Extract `WebhookController` from `ForemanWss`: webhook routing (`routePrEvent`, `routeCheckEvent`, `routeIssueEvent`, `forwardEvent`) registered via `webhooks.onAny()`; installation lifecycle delegated to `InstallationsController`
- Extract `ForemanWorkerController` from `ForemanWss`: all worker message business logic (hello handshakes, task assignment, reconnect, claim flow)
- Slim `ForemanWss` to pure transport: WS setup, ping/pong heartbeat, upgrade handling, `Map<type, handler>`-based message router
- `HttpServer` always receives a non-null `Webhooks` instance; dev mode uses `webhooks.receive()` to bypass signature verification instead of bypassing the registry

## Test plan

- [x] All 1905 existing tests pass unchanged (Supabase-requiring DB tests excluded as before)
- [x] TypeScript type check clean (`npx tsc --noEmit`)
- [x] ESLint clean (0 errors, pre-existing warnings only)
- [x] Test files updated to use new class locations (`ForemanWorkerController`, `WebhookController`, `WorkerMessenger`) — no test logic changed, only test harness updated

Closes #1064


src/foreman/controllers/webhook-controller.ts
```diff
@@ -0,0 +1,345 @@
...
+      ({ task, ref } = await this.routePrReviewEvent(p, evt));
+    } else if (name === "check_run" || name === "check_suite") {
+      ({ task, ref } = await this.routeCheckEvent(p, evt, name));
+    } else {
+      const issue = p.issue as R | undefined;
+      const issueNumber = numProp(issue, "number");
+      if (issueNumber !== null) {
+        ({ task, ref } = await this.routeIssueEvent(p, evt, issue!, issueNumber));
+      }
+    }
```
jasoncrawford: Instead of this cascading if statement, can we have a routing infrastructure where the WebhookController can register handlers, like the ApiController does with app.get()? I'm imagining something like `webhook.name("installation", (p) => { /* body of the routeInstallationEvent() method */ })`, instead of having routeInstallationEvent.

src/foreman/controllers/wss.ts
```diff
@@ -1,69 +1,82 @@
...
+  });
+  map.set("claim_task", async (workerId, _ws, msg) => {
+    await controller.handleClaimTask(workerId, msg as Extract<Wire.WorkerMessage, { type: "claim_task" }>);
+  });
+  map.set("worker_ready", async (workerId, _ws, _msg) => {
+    await controller.handleWorkerReady(workerId);
+  });
+  map.set("worker_reserved", async (workerId, _ws, _msg) => {
+    await controller.handleWorkerReserve(workerId);
+  });
```
jasoncrawford: Similar to my comment on the WebhookController: Instead of this, can we have the WorkerController register its own handlers for each type? Like `workerMessage.handle("worker_hello", (workerId, ws, msg) => { /* body of handleWorkerHello */ });`

src/foreman/controllers/wss.ts
```diff
@@ -169,82 +188,15 @@ export class ForemanWss {
         socket.destroy();
       }
     });
-
   }
 
+  /** Convenience method for tests and dev-mode: route an event directly without webhook verification. */
```
jasoncrawford: let's not have convenience methods for test/dev intruding on production code. do this a different way

src/foreman/controllers/wss.ts
```diff
@@ -258,615 +210,13 @@ export class ForemanWss {
     });
   }
 
+  /** Expose sendMsg for backwards compatibility with tests that spy on it. Delegates to messenger.send. */
```
jasoncrawford: can we not have backwards compatibility? just update the tests as needed

src/foreman/controllers/wss.ts
```diff
@@ -258,615 +210,13 @@ export class ForemanWss {
...
-   *
-   * 1. Unknown task (numeric taskId) → create placeholder, reclaim
-   * 2. Unknown task (non-numeric taskId) → cancel
-   * 3. Task belongs to a different repo → cancel
-   * 4. Complete task, same worker → reclaim for finalization
-   * 5. Complete task, different worker → cancel
-   * 6. Live task, different worker → cancel
-   * 7. Otherwise (live task, same or no worker) → reclaim
-   */
+  /** Proxy for integration tests that call handleAssignedHello directly. */
```
jasoncrawford: please no proxies just for tests. update the tests to call methods where they live

src/foreman/controllers/foreman-worker-controller.ts
```diff
@@ -0,0 +1,384 @@
...
+import { shortWorkerId } from "../../../shared/utils.js";
+import { getConfig } from "../../config.js";
+import type { BrunelConfig } from "../../config.js";
+
+type ForemanWorkerControllerOptions = {
+  config: Pick<BrunelConfig, "taskLabel" | "githubToken" | "githubApiUrl" | "workerSecret" | "pingIntervalMs">;
+  messenger: WorkerMessenger;
+};
+
+export class ForemanWorkerController {
```
jasoncrawford: Kind of a long class name; since the whole thing is the foreman, can we just call this WorkerController? It handles worker messages, that's pretty straightforward

jasoncrawford (commented):
This is going in a good direction. See detailed comments below.

Also, we now have two server modules in the controllers directory. Maybe move those up a level to src/foreman? Or if they should really be in a subdir, call it servers.

I'm ambivalent about where WorkerMessenger should go. It's not a controller, but it's kind of a controller helper object, so there are arguments for and against putting it in the controllers directory

src/foreman/controllers/webhook-controller.ts
jasoncrawford: Done — replaced the cascading if/else with a private `handlerMap` and an `on(name, handler)` registration method. Handlers for installation events are inlined in the constructor; named event types (pull_request, check_run, etc.) each call their existing route methods. Everything else falls through to the issue catch-all.

src/foreman/controllers/wss.ts
jasoncrawford: Done — moved the handler registration into `WorkerController`'s constructor. It fills a private `handlers` map (same keys as before) and exposes a `dispatch(workerId, ws, msg)` method. `wss.ts` now just calls `this.workerController.dispatch(workerId, ws, msg)`.

src/foreman/controllers/wss.ts
jasoncrawford: Removed. Tests that needed to fire webhook events now call `foremanWss.webhookController.handleEvent()` directly — the controller is exposed as a public `readonly` property on `ForemanWss`.

src/foreman/controllers/wss.ts
jasoncrawford: Removed, and tests updated directly.

src/foreman/controllers/wss.ts
jasoncrawford: Removed. The one test that called `foremanWss.handleAssignedHello()` now calls `foremanWss.workerController.handleAssignedHello()` — `workerController` is a public `readonly` property.

src/foreman/controllers/foreman-worker-controller.ts
jasoncrawford: Renamed to `WorkerController`. Also renamed the file from `foreman-worker-controller.ts` to `worker-controller.ts`.

jasoncrawford (commented):
Addressed all comments:

- Moved `http-server.ts` and `wss.ts` up to `src/foreman/` (out of controllers/)
- Renamed `ForemanWorkerController` → `WorkerController` (file: `worker-controller.ts`)
- `WorkerController` now registers its own message handlers in the constructor; `wss.ts` calls `dispatch()`
- `WebhookController` uses a private `handlerMap` + `on()` registration; no more cascading if/else
- Removed `routeEvent()`, `sendMsg()`, and `handleAssignedHello()` proxies from `ForemanWss`; tests updated to call `webhookController.handleEvent()` and `workerController.handleAssignedHello()` directly
- Left `WorkerMessenger` in controllers/ — accessible via `workerController.messenger` when needed from `wss.ts` or tests

src/foreman/controllers/webhook-controller.ts
```diff
@@ -0,0 +1,348 @@
...
+  private readonly assignWork: () => Promise<void>;
+  private readonly installationsController = new InstallationsController();
+  private readonly handlerMap = new Map<string, WebhookHandler>();
+
+  constructor({ webhooks, config, messenger, assignWork }: WebhookControllerOptions) {
+    this.config = config;
+    this.messenger = messenger;
+    this.assignWork = assignWork;
+
+    this.on("installation", async (p) => {
```
jasoncrawford: * Instead of doing this in the constructor, let's give each controller a `register` method, like ApiController?
* Instead of `on`, let's call it `handle` to make it distinct from event processing

src/foreman/controllers/webhook-controller.ts
```diff
@@ -0,0 +1,348 @@
...
+      } catch (err) {
+        log(`ERROR handling installation_repositories/${action}: ${fmtError(err)}`);
+      }
+    });
+
+    this.on("pull_request", (p, evt) => this.routePrEvent(p, evt));
+    this.on("pull_request_review", (p, evt) => this.routePrReviewEvent(p, evt));
+    this.on("pull_request_review_comment", (p, evt) => this.routePrReviewEvent(p, evt));
+    this.on("check_run", (p, evt) => this.routeCheckEvent(p, evt, "check_run"));
+    this.on("check_suite", (p, evt) => this.routeCheckEvent(p, evt, "check_suite"));
```
jasoncrawford: let's put the bodies of these handlers inline, like the ones above

src/foreman/controllers/worker-controller.ts
```diff
@@ -0,0 +1,418 @@
...
+    });
+    this.handlers.set("claim_task", async (workerId, _ws, msg) => {
+      await this.handleClaimTask(workerId, msg as Extract<Wire.WorkerMessage, { type: "claim_task" }>);
+    });
+    this.handlers.set("worker_ready", async (workerId) => {
+      await this.handleWorkerReady(workerId);
+    });
+    this.handlers.set("worker_reserved", async (workerId) => {
+      await this.handleWorkerReserve(workerId);
+    });
```
jasoncrawford: * put this in a `register` method
* call it `handle`
* put the bodies of the handlers inline

src/foreman/index.ts
```diff
@@ -19,9 +19,9 @@ const isMain = process.argv[1] === fileURLToPath(import.meta.url);
 if (isMain) {
   const config = await loadConfig(process.argv);
 
-  const webhooks = config.webhookSecret
-    ? new Webhooks({ secret: config.webhookSecret })
-    : null;
+  // Always create a Webhooks instance. In dev mode (no webhook secret), use a
+  // placeholder secret — receive() bypasses verification so the secret is unused.
+  const webhooks = new Webhooks({ secret: config.webhookSecret ?? "dev-mode-placeholder" });
```
jasoncrawford: I don't like putting dev-mode placeholders in production code. Can we do this a cleaner way?

jasoncrawford (commented):
Since we have HttpServer, ForemanWss, and AdminWss, let's make a new servers directory and put those three in there. See other detailed comments below.

src/foreman/controllers/webhook-controller.ts
```diff
@@ -0,0 +1,350 @@
...
+      } catch (err) {
+        log(`ERROR handling installation_repositories/${action}: ${fmtError(err)}`);
+      }
+    });
+
+    this.handle("pull_request", (p, evt) => this.routePrEvent(p, evt));
+    this.handle("pull_request_review", (p, evt) => this.routePrReviewEvent(p, evt));
+    this.handle("pull_request_review_comment", (p, evt) => this.routePrReviewEvent(p, evt));
+    this.handle("check_run", (p, evt) => this.routeCheckEvent(p, evt));
+    this.handle("check_suite", (p, evt) => this.routeCheckEvent(p, evt));
```
jasoncrawford: can we just inline these methods?

src/foreman/controllers/worker-controller.ts
```diff
@@ -0,0 +1,424 @@
...
+    this.handle("claim_task", async (workerId, _ws, msg) => {
+      await this.handleClaimTask(workerId, msg as Extract<Wire.WorkerMessage, { type: "claim_task" }>);
+    });
+    this.handle("worker_ready", async (workerId) => {
+      await this.handleWorkerReady(workerId);
+    });
+    this.handle("worker_reserved", async (workerId) => {
+      await this.handleWorkerReserve(workerId);
+    });
```
jasoncrawford: can we inline all these methods?

jasoncrawford (commented):
Thanks for the feedback! Two different answers for the two files:

**webhook-controller.ts lines 80–84:** The routing methods (, , , ) are all called directly by  — there are ~30 call sites that test them in isolation. Inlining their bodies into the lambdas here would require moving all that coverage to integration-style tests through the full `handleEvent` → dispatch path, which would make each test case significantly harder to write and less targeted. I'd prefer to keep them as named public methods for that reason.

**worker-controller.ts lines 37–57:** The `handle*` methods here are only called from `register()` and aren't tested directly, so they *could* be inlined. The wrapping lambdas are thin and already read almost like one-liners. The reason I kept them as separate methods is that each is 20–50 lines, so inlining all seven would make `register()` ~200+ lines — a lot to scan to find the dispatch wiring vs. the business logic. Happy to inline them if you'd prefer that shape, just flagging the trade-off.

jasoncrawford:
Correction to my previous comment (shell ate the backtick-quoted names):

**webhook-controller.ts lines 80–84:** The routing methods `routePrEvent`, `routePrReviewEvent`, `routeCheckEvent`, and `routeIssueEvent` are all called directly by `foreman.event-handlers.test.ts` — there are ~30 call sites that test them in isolation. Inlining their bodies into the lambdas would require moving all that coverage to integration-style tests through the full `handleEvent` → dispatch path, making each test case significantly harder to write and less targeted. Keeping them as named public methods preserves that granularity.

**worker-controller.ts lines 37–57:** The `handle*` methods are only called from `register()` and are not tested directly, so they *could* be inlined. The wrapping lambdas are already thin one-liners. The reason I kept them separate is that each is 20–50 lines, and inlining all seven would make `register()` ~200+ lines — hard to scan to find the dispatch wiring vs. the business logic. Happy to inline them if you prefer that shape.

jasoncrawford (commented):
OK. Different approach: can we route by convention instead of configuration?

In WorkerController, we could discard the map, register(), and handle() completely. Instead, any message named foo_bar gets mapped to a handler called handleFooBar().

In WebhookController, we could do something similar: a webhook event named foo_bar goes to handleFooBarEvent, or if we want to keep the "route" naming scheme, routeFooBarEvent. This would require renaming some methods, and some method names would get long, like handlePullRequestReviewCommentEvent. But we would gain consistency and predictability. Again, we could drop the handlerMap, register(), and handle().

Issue #1085, opened 2026-05-09
Unpause events as soon as we are waiting for tasks


jasoncrawford:
When you complete a task (`/worker:complete`) and then wait for tasks, the status bar says "Events paused" while waiting for tasks.

We should unpause events as soon as we are waiting, so it doesn't say that.

PR #1091, opened 2026-05-09
refactor: consolidate idle-state invariants into _setIdleState()

### LINKED ISSUE #1089: Status bar shows 'no current task' instead of 'waiting for task' after completing a task
## Description

When you complete a task (`/worker:complete`) and choose "Wait to be assigned the next task", the status bar shows **"no current task"** instead of **"waiting for task"**.

## Root cause

`promptAfterTaskComplete()` calls `sendWorkerReady()` to opt back into auto-assignment, but never calls `transitionToIdle()` (which is what sets `agentStatus.workerReady = true`). The renderer shows "waiting for task" only when `workerReady` is `true`, so it falls through to "no current task".

Compare with `/worker:start`, which correctly calls both:
```typescript
this.sendWorkerReady();
this.transitionToIdle();  // <-- sets workerReady: true
```

The fix is to also update `agentStatus.update({ workerReady: true })` in the "Wait for next task" branch of `promptAfterTaskComplete()` (either directly or by calling `transitionToIdle()` without its redundant print).

## Related

Noticed while reviewing the fix for #1085 (events paused showing in the same state).

Labels: polish, worker

jasoncrawford:
## Summary

- Extracts `_setIdleState()` from `transitionToIdle()` to own the invariants that must hold on every entry to the waiting state: clear `_eventsPaused`, sync status, set `workerReady: true`
- `transitionToIdle()` becomes `transitionToRegistered()` + `_setIdleState()` + print — used by reconnect paths
- `promptAfterTaskComplete()` calls `_setIdleState()` directly since the connection is already registered (calling `transitionToRegistered()` there would prematurely flush the message buffer during reconnect)
- Updates CLAUDE.md to reflect the new canonical entry point

## Why

Previously, `transitionToIdle()` bundled connection-state management with task-state management. `promptAfterTaskComplete()` couldn't call it without risking an incorrect buffer flush (as caught by existing tests). This left the idle-state invariants scattered: PR #1085 added the `_eventsPaused` reset directly to `promptAfterTaskComplete()`, but `workerReady` was still never set, causing the status bar to show "no current task" instead of "waiting for task" after completing a task (issue #1089).

The refactor makes `_setIdleState()` the single place to update. Any future path that enters the waiting state calls it and automatically gets all invariants.

## Test plan

- New test: `option 0 (wait for next task): sets workerReady to true so status bar shows 'waiting for task'`
- Existing test passes: `option 0 (wait for next task): clears eventsPaused so status bar does not show 'Events paused'`
- Existing buffering tests continue to pass (the buffer-flush regression is prevented by the refactor)

Closes #1089


CLAUDE.md
```diff
@@ -117,7 +117,7 @@ See **`docs/type-system.md`** for the full design. In brief: one server model cl
...
+- **Worker state model**: `WorkerController` has three explicit states — stopped (REPL prompt `>`), waiting (connected, no task), and active (task assigned). `_deactivate()` is the canonical entry for the stopped state — every deactivation path (`stop()`, fatal `foreman_error`) must call it; it sets `_isActive`, `workerModeActive`, `connectionStatus`, and `reconnectAt` atomically. `_setIdleState()` owns the idle-state invariants (clears `_eventsPaused`, syncs status, sets `workerReady: true`) and must be called by every code path that enters the waiting state. `transitionToIdle()` = `transitionToRegistered()` + `_setIdleState()` + print, used by reconnect paths; `promptAfterTaskComplete()` calls `_setIdleState()` directly since the connection is already registered. The waiting state has two sub-states tracked by `_isReserved`: ready (auto-assignable, status bar shows "waiting for task") and reserved (not auto-assignable, status bar shows "no current task"). The `workerReady` field on `AgentStatus` carries this distinction to the renderer. `WorkspaceController` registers all event listeners once in the constructor; do not add `workspace.on(…)` calls anywhere else.
```
jasoncrawford: Too much implementation detail here. Let's keep CLAUDE.md a high-level guide. More detailed info should be in comments in the code, if it's not evident from reading the code itself.

PR #1126, opened 2026-05-13
feat: add /worker:resume command to take over dead worker identity

### LINKED ISSUE #1123: Add `/worker:resume <prefix>` to take over a dead worker's session
## Background

When a worker process dies unexpectedly (terminal killed, process crash, accidental \`pkill\`, etc.), there is currently no way to resume it. The Claude conversation history is preserved in \`~/.claude/projects/\` as a JSONL session file, and the workspace git clone still exists at \`~/.brunel/workers/<agent-id>/\`. But starting a new worker creates a fresh identity, fresh workspace, and a fresh Claude session — losing all that context and requiring the foreman to re-assign from scratch.

This issue tracks a \`/worker:resume <prefix>\` command that lets a new worker process take over a dead worker's identity, workspace, task assignment, and Claude session — as if the crash never happened.

## Design overview

The worker finds the existing workspace directory by prefix match, derives the dead worker's full agent ID from the directory name, assumes that identity, and connects to the foreman with a new \`"resume"\` hello status. The foreman looks up the dead worker by ID, confirms it is disconnected, and replies with whatever state it has recorded (assigned task, ready, reserved). The resuming worker handles the \`hello_ack\` through the existing path, additionally restoring the Claude session ID so the conversation picks up mid-thread.

As far as the foreman is concerned, this is indistinguishable from the original process reconnecting — except that the \`"resume"\` status tells it _"I've lost my state, tell me what I have"_ rather than the worker asserting a task ID as in the normal reconnect path.

## Wire protocol

**\`shared/wire.ts\`** — add \`"resume"\` to \`WorkerHello.status\`:

\`\`\`typescript
// before
status: "ready" | "reserved" | "assigned";

// after
status: "ready" | "reserved" | "assigned" | "resume";
\`\`\`

No new fields needed. The resuming worker sends its assumed \`workerId\` (the dead worker's full agent ID) in the existing field. All other fields (\`repo\`, \`version\`, \`protocolVersion\`, auth) are sent as normal.

Bump \`PROTOCOL_VERSION\`.

## Foreman changes

**\`src/foreman/controllers/worker-controller.ts\`** — add a \`handleResumeHello()\` branch routed from \`handleWorkerHello()\`:

1. Auth validation runs as normal (same as all other paths).
2. Check the in-memory worker registry for \`workerId\`. If the worker is **currently connected**: send a fatal \`foreman_error\` — _"Worker is currently connected; cannot resume a live worker."_
3. Look up the worker in the DB via \`Worker.get(workerId)\`. If **not found**: send a fatal \`foreman_error\` — _"No worker found with id \`<id>\`."_
4. Validate that the \`repo\` field matches the worker's recorded repo. Mismatch → fatal error.
5. Check for an active task via \`Task.getByWorker(workerId)\`:
   - **Task exists and is not complete**: re-register the worker, reclaim the task (same \`reclaimWorker()\` call as the reconnect path), send \`hello_ack { status: "assigned", ... }\`.
   - **No task, or task is complete**: register fresh via \`_registerFresh()\` (which reverts any stale task), send \`hello_ack { status: "ready" }\`.

Event replay (\`replayMissedEvents\`) is **skipped** in this initial implementation — the resuming worker does not know \`lastSeenEventSeqId\` since it has no prior state. New events arriving after reconnect are forwarded normally.

### 1. \`AgentStatus.setAgentId()\` — assume the dead worker's identity

\`AgentStatus.agentId\` is currently immutable after construction. Add a \`setAgentId(id: string)\` method so \`/worker:resume\` can override it before connecting to the foreman.

### 2. \`Workspace.attach()\` — use an existing directory without cloning

\`Workspace.create()\` always clones and runs \`npm install\`. Add a parallel \`Workspace.attach()\` method for the resume case:
- Verifies the directory exists and contains a \`.git\`.
- Re-applies git auth configuration (same as \`create()\` does post-clone).
- Writes \`.brunel.lock\` (marks the workspace as owned by this process).
- Does **not** clone or run \`npm install\`.

### 3. \`WorkspaceController.attachExisting(dir)\` — drive the attach lifecycle

Add a method that creates a \`Workspace\` pointed at the existing directory, calls \`attach()\`, and \`chdir()\`s into it. The \`WorkspaceController\` then owns teardown for this workspace on task completion or exit, same as a normally created workspace.

### 4. \`QueuedPrompt\` — carry the Claude session ID

Add an optional \`resumeSessionId?: string\` field to the \`QueuedPrompt\` type. When \`hello_ack { status: "assigned" }\` arrives in the resume path, look up the most recent JSONL session file in \`~/.claude/projects/\` for the workspace directory and populate this field on the first queued prompt.

In \`index.ts\`, before running a prompt, add one line alongside the existing \`fresh\` check:

\`\`\`typescript
if (item.resumeSessionId) sessionId = item.resumeSessionId;
else if (item.fresh) sessionId = undefined;
\`\`\`

This is the only change needed to \`index.ts\`. The command itself lives entirely in \`WorkerController.registerCommands()\` — it stashes the found session ID internally, and it rides along on the first queued prompt via \`resumeSessionId\`. No callback or event is needed to bridge \`sessionId\` access across the boundary.

### 5. \`/worker:resume <prefix>\` command

Registered in \`WorkerController.registerCommands()\`. The handler:

1. Globs \`<workspaceDir>/<prefix>*/\` (using the configured \`workspaceDir\`, defaulting to \`~/.brunel/workers\`).
   - 0 matches → error: _"No worker workspace found matching '\`<prefix>\`'."_
   - 2+ matches → error: _"Ambiguous — multiple workspaces match '\`<prefix>\`': \`<list>\`."_
2. Extracts the full agent ID from the directory name (the directory name _is_ the agent ID).
3. Calls \`agentStatus.setAgentId(fullAgentId)\`.
4. Calls \`workspaceController.attachExisting(dir)\` to take over the workspace.
5. Finds the most recent \`.jsonl\` file (by mtime) in the corresponding \`~/.claude/projects/\` directory and stashes the session UUID as \`_pendingResumeSessionId\`.
6. Initiates the worker connection with \`status: "resume"\` instead of the normal \`"ready"\`.

When \`hello_ack { status: "assigned" }\` arrives, the stashed session ID is attached to the first queued prompt via \`resumeSessionId\`. \`_pendingResumeSessionId\` is cleared after use.

If the foreman replies with \`cancelled\` or \`ready\`, the worker becomes idle in the attached workspace with no Claude session to restore.

## Session file lookup

The Claude SDK stores sessions at:
\`\`\`
~/.claude/projects/<encoded-path>/<session-uuid>.jsonl
\`\`\`
where \`<encoded-path>\` is the workspace directory with \`/\` replaced by \`-\`. For example, \`/home/node/.brunel/workers/harold-6ee65735-…\` maps to \`-home-node--brunel-workers-harold-6ee65735-…\`.

Pick the most recently modified \`.jsonl\` file in that directory. If none exists, proceed without a session ID (the agent starts the conversation fresh but still reclaims the task and workspace).

## Tests

- **Wire**: \`PROTOCOL_VERSION\` is bumped; \`"resume"\` is accepted in \`WorkerHello.status\`.
- **Foreman unit tests** (\`tests/foreman.hello-handlers.test.ts\`): resume with connected worker → fatal error; resume with unknown worker → fatal error; resume with disconnected worker that has active task → \`assigned\` ack + task reclaim; resume with disconnected worker with no task → \`ready\` ack.
- **\`AgentStatus\`**: \`setAgentId()\` updates \`agentId\`; \`workerId\` sent in hello reflects the overridden value.
- **\`Workspace.attach()\`**: existing directory → skips clone, writes lock, configures auth; missing directory → throws; directory without \`.git\` → throws.
- **\`/worker:resume\` command**: prefix resolution (no match, ambiguous, unique); correct agent ID extracted from directory name; correct \`status: "resume"\` sent in hello.

jasoncrawford:
## Summary

- Adds `/worker:resume <prefix>` command: finds a workspace by directory name prefix, assumes the dead worker's agentId, attaches to the existing workspace, and connects with `status: "resume"` so the foreman can re-assign any in-progress task
- Adds `handleResumeHello` in the foreman: validates the worker isn't live, exists in DB, and matches the repo; reclaims active tasks (sending both `hello_ack { assigned }` and `task_assigned`) or registers fresh (sending `hello_ack { ready }`)
- Attempts to locate and resume the dead worker's last Claude session ID so conversation history is preserved
- Bumps `PROTOCOL_VERSION` to 2 (wire protocol now includes `"resume"` status)

## Changed files

- `shared/wire.ts` — adds `"resume"` to worker_hello status union, bumps `PROTOCOL_VERSION` to 2
- `src/agent/models/agent-status.ts` — adds `setAgentId()` for identity takeover
- `src/agent/models/workspace.ts` — adds `attach()` for joining existing dirs; exposes `repoUrl`
- `src/foreman/models/worker.ts` — exposes `repoId` for repo-match validation
- `src/foreman/controllers/worker-controller.ts` — routes `status: "resume"` to `handleResumeHello()`
- `src/agent/controllers/workspace-controller.ts` — mutable `_workspace`, adds `attachExisting()`
- `src/agent/controllers/worker-controller.ts` — `/worker:resume` command, `_isResuming` flag, `resumeSessionId` in queued prompts
- `src/agent/index.ts` — honours `resumeSessionId` in `drainPendingPrompts`
- `tests/worker.resume.test.ts` — new test suite covering prefix resolution, agent ID takeover, and resume hello
- `CLAUDE.md` — documents the new resume flow

## Test plan

- [ ] All existing tests pass (`npm test`)
- [ ] New `tests/worker.resume.test.ts` covers: no-match, ambiguous-match, exact-match prefix resolution; agentId set to directory name; `status: "resume"` in first hello; subsequent reconnects use `"ready"` not `"resume"`; `PROTOCOL_VERSION >= 2`; `"resume"` is valid in TypeScript union

Closes #1123


shared/wire.ts
```diff
@@ -4,7 +4,7 @@
 export type TaskStatus = "pending" | "assigned" | "pushed" | "merged" | "closed" | "complete" | "blocked";
 
 /** Current foreman ↔ worker wire protocol version. Increment when making incompatible changes. */
-export const PROTOCOL_VERSION = 1;
+export const PROTOCOL_VERSION = 2;
```
jasoncrawford: Will this now reject all clients that are speaking protocol 1? Not really necessary since the only thing new that is added is the "resume" that the client can send?

src/agent/controllers/worker-controller.ts
```diff
@@ -745,6 +752,81 @@ export class WorkerController extends EventEmitter {
...
+        }
+
+        // Find workspace directories matching the prefix
+        let matches: string[] = [];
+        try {
+          matches = fs.readdirSync(workspaceDir, { withFileTypes: true })
+            .filter(e => e.isDirectory() && e.name.startsWith(prefix))
+            .map(e => e.name);
+        } catch {
+          this.display.print(c.boldRed("No worker workspace found matching the prefix."));
```
jasoncrawford: I think this should say there was an error looking for the workspace, and what the error was

src/agent/controllers/worker-controller.ts
```diff
@@ -745,6 +752,81 @@ export class WorkerController extends EventEmitter {
...
+          return undefined;
+        }
+        if (matches.length > 1) {
+          this.display.print(c.boldRed(`Ambiguous prefix "${prefix}": matches ${matches.join(", ")}`));
+          return undefined;
+        }
+
+        const agentId = matches[0];
+
```
jasoncrawford, line 796: One thing we should maybe check here is whether the worker is, in fact, running. We could check the pid file and look for that process, the way the `prune` command does. If process still running, error.

jasoncrawford (commented):
If there is a task, and the foreman send a hello_ack with task_assigned, will the worker create a new task-assignment prompt? That's not actually what we want in this case. We want the worker to just resume the session where it left off.

Actually, will the agent SDK just resume a session without a prompt? Or do we need to give it a prompt? If so then we need to come up with a special resumption prompt.

(This makes me wonder if the foreman should actually send a task_resumed response or something similar to clearly make it different)

jasoncrawford (commented):
Thanks for the review! Here's what I did:

**Resume prompt** — You're right that re-sending the full task assignment prompt into an existing Claude session would be confusing. The SDK does require a prompt on every call (there's no "resume silently" mode), so I added a `buildResumePrompt` that just says: *"You are resuming work on GitHub issue #N: '…'. Please review the conversation history to recall the current state of the work, and continue from where the previous session left off."* When `resumeSessionId` is set, the `task_assigned` handler now uses this short prompt and passes `fresh=false` (no session reset). (6e27180)

**`task_resumed` vs `task_assigned`** — I kept the foreman sending `task_assigned` for now. The distinction is handled entirely in the worker: if `_pendingResumeSessionId` is set, it knows to use the resume prompt. A separate `task_resumed` message would make the intent more explicit, but it's a larger wire-protocol change and doesn't add behaviour we need today. Worth reconsidering if the foreman ever needs to know whether the worker is fresh vs resumed.

**PID liveness check** (line 796) — Added. Before taking over, the command now reads `.brunel.lock` and calls `process.kill(pid, 0)` (the same check the `prune` command uses). If the process is still alive, it prints an error and bails. (6e27180)

**Error message** (line 782) — Fixed: the `catch` block now includes the actual error via `fmtError(err)`. (6e27180)

**`PROTOCOL_VERSION`** (line 7) — Agreed: `"resume"` is client→server only, so old workers are still compatible. Reverted to 1 and removed the `>= 2` assertion from tests. (6e27180)

PR #1135, opened 2026-05-14
fix: detach workspace on foreman rejection during /worker:resume

### LINKED ISSUE #1134: attach() writes new PID to lock file before checking the old one, causing workspace destruction on resume failure
## Bug

\`Workspace.attach()\` overwrites \`.brunel.lock\` with the new process PID and sets \`isCreated = true\` unconditionally, without first checking whether the existing lock PID belongs to a live process.

If the resume flow fails after \`attach()\` (e.g. the foreman rejects the connection because the original worker is still registered as connected), the lock file now holds the new REPL's own PID. On the next attempt, the guard check in \`WorkerController\` reads the lock back, finds the REPL's own PID alive, and refuses with "still running. Stop it first." If the user follows that instruction and kills the reported PID, they kill their own REPL — and because \`isCreated\` was already set to \`true\`, the REPL's exit handler calls \`workspace.destroy()\`, deleting the workspace directory and any unpushed work.

## Root cause

\`attach()\` (workspace.ts) and the pre-call guard (worker-controller.ts) are split: the guard checks the lock before calling \`attach()\`, but \`attach()\` overwrites the lock without rechecking. Any failure between the guard check and a successful foreman connection leaves the lock in a corrupt state (owned by the REPL), making the workspace unclaimed by any resumed worker but also appearing live to future attempts.

## Fix

Move the lock check inside \`attach()\` itself, before writing the new PID. This makes the check-and-claim atomic and ensures \`isCreated\` is never set if the workspace is already live. The guard in \`WorkerController\` can stay as a fast-path, but \`attach()\` is now the authoritative enforcer.

\`\`\`typescript
// check existing lock BEFORE writing new one
const lockPath = path.join(this.dir, ".brunel.lock");
if (fs.existsSync(lockPath)) {
  const existing = parseInt(fs.readFileSync(lockPath, "utf8").trim(), 10);
  if (!isNaN(existing) && isProcessAlive(existing)) {
    throw new Error(\`Workspace is owned by process \${existing}, which is still running\`);
  }
}
await this._configureAuth();
fs.writeFileSync(lockPath, String(process.pid));
// ...
this.isCreated = true;
\`\`\`

This fix has been applied.

jasoncrawford:
## Summary

- `Workspace.attach()` now checks the existing lock PID **before** overwriting it, throwing (without setting `isCreated`) if the workspace is still live
- `WorkerController` tracks `_attachedViaResume` and calls `workspace.detach()` on a fatal `foreman_error`, releasing the lock without deleting the directory

## Root cause

When `/worker:resume` ran and the foreman rejected the connection (e.g. "worker currently connected"), `attach()` had already written the REPL's own PID to the lock file and set `isCreated = true`. A second resume attempt then read the REPL's own PID from the lock, found it alive, and refused with "still running. Stop it first." — causing the user to kill their own REPL, whose exit handler then destroyed the workspace and all unpushed work.

## Test plan

- `Workspace.attach()` throws without overwriting the lock when the existing PID is alive
- `Workspace.attach()` proceeds normally when the existing lock PID is dead
- `Workspace.detach()` clears `isCreated`, removes the lock, leaves the directory intact
- `WorkerController`: fatal foreman error removes the lock (workspace detached)
- `WorkerController`: a second `/worker:resume` succeeds after a fatal rejection

Fixes #1134

### DIFF
--- a/src/agent/controllers/worker-controller.ts
+++ b/src/agent/controllers/worker-controller.ts
@@ -797,24 +797,18 @@ export class WorkerController extends EventEmitter {
 
         const agentId = matches[0];
 
-        // Refuse to take over a workspace whose worker process is still alive.
-        const lockPath = path.join(workspaceDir, agentId, ".brunel.lock");
-        if (fs.existsSync(lockPath)) {
-          const pid = parseInt(fs.readFileSync(lockPath, "utf8").trim(), 10);
-          if (!isNaN(pid)) {
-            try {
-              process.kill(pid, 0); // throws ESRCH if not running
-              this.display.print(c.boldRed(`Worker ${agentId} (PID ${pid}) is still running. Stop it first.`));
-              return undefined;
-            } catch { /* process is gone — safe to resume */ }
-          }
-        }
-
         // Take on the dead worker's identity
         this.agentStatus.setAgentId(agentId);
 
-        // Attach to the existing workspace directory
-        await this.workspaceController!.attachExisting(agentId, this.options?.githubToken);
+        // Attach to the existing workspace directory. attach() enforces the
+        // live-lock check and throws if the workspace is still owned by a live
+        // process — no need to duplicate that check here.
+        try {
+          await this.workspaceController!.attachExisting(agentId, this.options?.githubToken);
+        } catch (err) {
+          this.display.print(c.boldRed(`Cannot resume: ${fmtError(err)}`));
+          return undefined;
+        }
         this._attachedViaResume = true;
 
         // Look up the last Claude session for this workspace
@@ -1100,7 +1094,7 @@ export class WorkerController extends EventEmitter {
         // If we attached a workspace for this resume but the foreman rejected us,
         // detach (release the lock) without destroying the directory.
         if (this._attachedViaResume) {
-          this.workspaceController?.workspace?.detach();
+          this.workspaceController?.onDetach();
           this._attachedViaResume = false;
         }
         this._deactivate();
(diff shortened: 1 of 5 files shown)


src/agent/models/workspace.ts
```diff
@@ -208,6 +215,17 @@ export class Workspace extends EventEmitter {
     return removed;
   }
 
+  /**
+   * Undo an attach() without deleting the directory. Removes the lock file and
+   * clears isCreated so the workspace is no longer considered owned by this process.
+   * Used when the foreman rejects a resume attempt after attach() has already run.
+   */
+  detach(): void {
+    const lockPath = path.join(this.dir, ".brunel.lock");
```
jasoncrawford: this now exists in 2-3 places, if it gets out of sync that's a bug… DRY?

src/agent/controllers/worker-controller.ts
```diff
@@ -1092,6 +1097,12 @@ export class WorkerController extends EventEmitter {
 
     if (msg.type === "foreman_error") {
       if (msg.fatal) {
+        // If we attached a workspace for this resume but the foreman rejected us,
+        // detach (release the lock) without destroying the directory.
+        if (this._attachedViaResume) {
+          this.workspaceController?.workspace?.detach();
```
jasoncrawford: should we also change pwd? get ourselves out of the workspace if we're detaching. not sure where that belongs, wherever else we cwd

src/agent/models/workspace.ts
jasoncrawford: Fixed — removed the pre-guard from WorkerController entirely. `Workspace.attach()` is now the single enforcer; the caller just wraps it in a try/catch and prints the error.

src/agent/controllers/worker-controller.ts
jasoncrawford: Yes — added `WorkspaceController.onDetach()` which calls `workspace.detach()` and then `process.chdir(workspace.originalCwd)`. The fatal error handler now calls `workspaceController?.onDetach()`. Test also asserts `process.cwd()` is restored.

PR #1145, opened 2026-05-17
feat: settings namespace with verbose and think-out-loud controls

### LINKED ISSUE #708: Settings improvements
Make more things into settings changeable at runtime, e.g., verbose, think-out-loud, permisisons (cf #416)

Group these commands under a `settings` namespace? `/settings:verbose` for instance; again it would help to have aliases here so you can still type e.g. `/model` and it matches `/settings:model` (see #705)

Maybe a `/settings` command that displays all settings and their current values, lets you arrow down through a menu of them, if you hit enter it acts as if you invoked that specific command

All this should live in the settings module, which should register its own commands in the namespace

jasoncrawford:
## Summary

- **Settings namespace**: Moved `/model`, `/effort`, `/permissions` into the `settings:` namespace (`/settings:model`, `/settings:effort`, `/settings:permissions`). The old short forms (`/model`, `/effort`) still work via suffix matching.
- **New settings commands**: Added `/settings:verbose` (on/off) and `/settings:think-out-loud` (on/off/default). Accept common aliases — `on`/`yes`/`y` for true, `off`/`no`/`n` for false, case-insensitive.
- **`/settings` overview command**: Shows all settings with current values; Tab cycles through valid values for the highlighted setting (applying changes immediately); Enter opens the sub-picker for that setting; Escape cancels.
- **Settings model**: Now owns `verbose` and `thinkOutLoud` (seeded from config at startup). `Display` reads from `Settings` so runtime changes take effect without restart. `BrunelConfig.thinkOutLoud` is now `boolean | undefined` (undefined = default = same as verbose).

## Test plan

- [x] All 2010 existing unit tests pass (64 test files)
- [x] ESLint: 0 errors (3 pre-existing warnings in unrelated code)
- [x] New TDD tests cover: `Settings.verbose`, `Settings.thinkOutLoud`, `Settings.effectiveThinkOutLoud`, `pickVerbose` (direct set + picker, all aliases, case-insensitive), `pickThinkOutLoud` (direct set + picker, all aliases including `default`)
- [x] Updated tests for new `settings:` namespace in dispatch, autocomplete, slash resolution, and command registry tests

Closes #708


jasoncrawford:
In the settings menu:

* when an item is highlighted, all possible values should be shown in grey, with the current selection highlighted in white
* shift-tab should rotate backwards through the values, opposite of tab
* right and left arrow should also rotate through the values
* there should be a Done option at the end that exits the menu without making further changes

jasoncrawford:
* The model setting should also show all the available model values, like the other settings.
* Display bug: When I first start cycling through the options on any setting, all the text jumps up a line, and there are two "Done" lines.
* Another display bug: when I hit Done, everything disappears and I get a new status bar, rendered a few lines above the previously-rendered status bar (so now I see two status bars).

jasoncrawford:
The "Model" option has two "default" values and they are both highlighted when I arrow through them?

jasoncrawford:
Change of spec:

* Enter should not invoke a specific menu. It should just be like Done and complete the settings command. Done can be removed.
* When menu is first rendered, there should be a heading selected: "Arrows to choose settings, Enter when done". None of the specific options are highlighted/expanded until you arrow down to them.
* When enter is pressed, command exits but it does not erase the menu. All options are un-highlighted and "closed" to show the current settings values.

src/agent/index.ts
```diff
@@ -216,34 +216,19 @@ export class BrunelAgent {
...
-        await this.settingsController.pickPermissions(
-          args,
-          (opts, idx) => this.picker.pick(opts, { currentIdx: idx, escapable: true }),
-        );
-      },
-    });
+    const pickFn = (opts: string[], idx: number) =>
+      this.picker.pick(opts, { currentIdx: idx, escapable: true });
+    const settingsPickFn = (
+      entries: import("./views/picker.js").SettingsMenuEntry[],
```
jasoncrawford: no-inline-import

src/agent/index.ts
```diff
@@ -216,34 +216,19 @@ export class BrunelAgent {
...
-      handler: async (args) => {
-        await this.settingsController.pickPermissions(
-          args,
-          (opts, idx) => this.picker.pick(opts, { currentIdx: idx, escapable: true }),
-        );
-      },
-    });
+    const pickFn = (opts: string[], idx: number) =>
+      this.picker.pick(opts, { currentIdx: idx, escapable: true });
+    const settingsPickFn = (
```
jasoncrawford: do these functions need to exist as arguments? is this another injection thing for tests? if so then do tests a different way, to simplify production code.

jasoncrawford:
Fixed in the latest commit. Both issues stemmed from the same root:

- **Inline import** (line 222): the `import(...).SettingsMenuEntry[]` type annotation existed only because `settingsPickFn` was a local lambda needing the type — it's gone now.
- **Wrapper lambdas** (lines 219–221): these existed solely for test injection. Moved the picker into the `SettingsController` constructor (`private readonly picker: SettingsPicker`) so the controller calls `this.picker.pick()` and `this.picker.pickSettingsMenu()` directly. Tests now inject a mock picker via the constructor instead.

